Issue #3, opened 2026-07-13
#04: feat: FastAPI base with minimal endpoints, async DB connection, and structured logging


pradyumna-001:
## Overview
Implement the foundational FastAPI application skeleton for **FinAgent**. This includes configuring application lifespans for clean database connections, creating standard/multi-tenant endpoints, and setting up context-aware tracing middleware.

## Requirements & Scope
* **Application Lifespan**: Set up async database initialization on startup and safe connection pool cleanup (`engine.dispose()`) on shutdown.
* **Structured Tracing Middleware**: Build a middleware layer leveraging `contextvars` to dynamically append `pipeline_run_id` and `morning_note_id` to log records across all active routes.
* **Database Session Routing**: Ensure sessions handle raw SQL/ORM transactions dynamically and support Row-Level Security injections via transaction context variables.

## Target Endpoints
* **`GET /health`**
  - Returns: `{"status": "ok", "db": "ok", "redis": "ok", "age": "ok"}`
  - Logic: Performs a real `SELECT 1` ping verification against PostgreSQL.
* **`POST /pipeline/trigger`**
  - Returns: `202 Accepted` status including a unique `pipeline_run_id`.
  - Logic: Dispatches ingestion pipeline tasks asynchronously.
* **`GET /morning-notes`**
  - Returns: Isolated morning note records.
  - Constraint: Requires a valid `gestor-id` in the request header to satisfy PostgreSQL RLS requirements. Rejects missing headers with a `400 Bad Request`.

## Deliverables & Acceptance Criteria
- [ ] Endpoints must respond successfully via local `curl` triggers.
- [ ] The logging format automatically inherits tracing IDs from request headers or runtime triggers.
- [ ] Requests to `/morning-notes` without a `gestor-id` header are explicitly rejected.

PR #2, opened 2026-07-13
feat: implement database models, async alembic migrations, and rls po…

pradyumna-001:
## Overview
This PR fully implements the structural data layer for **FinAgent**, introducting the declarative database models, an asynchronous migration workflow with Alembic, and Row-Level Security for multi-tenant isolation.

Closes #1 

## Changes Included
* **SQLAlchemy Models (`app/db/models.py`)**: Defined multi-tenant schema.
*  **Alembic Configuration (`alembic.ini` & `env.py`)**: Initialized a complete `asyncpg` migration pipeline pointing to the local `postgresql_vector` target.
* **Vector Embeddings**: Configured vector dimension metrics to store AI-generated embeddings.
* **Database Security (RLS)**: Appended transactional DDL overrides to ensure runtime tenant isolation based on `manager_id`

## Verification & Testing
Run automated autogeneration and upgraded the database to head locally:
```powershell
uv run alembic upgrade head

PR #30, opened 2026-07-24
feat(issue-03): Closes #1 — Models, migrations, RLS, and migration tests

### LINKED ISSUE #1: #03: SQLAlchemy Models and Migrations
### Objective
Implement the database schema, indexes, and row-level security (RLS) tracking profiles on the relational tier (postgres_vector).

### Tasks
- [ ] Create SQLAlchemy model mappings for Manager, Company, Portfolio, MorningNote, and Recommendation
- [ ] Configure async Alembic migration environment
- [ ] Generate initial database migration footprint
- [ ] Inject composite index (gestor_id, empresa_id, data) and partial evaluation state index into the migration step
- [ ] Inject HNSW vector cosine operations index on morning note embeddings
- [ ] Author explicit PostgreSQL Row Level Security (RLS) tenant isolation policies

### Deliverables
Successful execution of `alembic upgrade head` provisioning all models, optimized indexes, and data isolation guards.

pradyumna-001:
# feat: Issue #03 — Models, migrations, indexes, RLS, and migration tests

Closes #1 (data layer rebuild; supersedes PR #2).

> **Rebuild context:** This branch rebuilds the data layer from issue #03
> forward. PR #2 carried the first-pass implementation; this PR replaces
> it with a complete, tested schema + RLS scaffolding.

## What this PR does

- **SQLAlchemy 2.x typed models** in `finagent/app/db/models.py`: six
  declarative classes (`Manager`, `Company`, `Portfolio`,
  `PortfolioHolding`, `MorningNote`, `Recommendation`), all using
  `Mapped[T] = mapped_column(...)` and a shared `Base(DeclarativeBase)`.
- **Async Alembic migrations** in `finagent/alembic/`: seven revisions
  from initial schema through RLS enablement, all runnable against an
  asyncpg target via `alembic upgrade head`.
- **Indexes**: composite `(manager_id, company_id, note_date)` on
  `morning_notes`, partial index on `morning_notes(status='completed')`,
  HNSW index on the `embedding halfvec(2048)` column.
- **RLS**: `morning_notes` enabled with a `manager_id` policy. RLS on
  the other five tables is intentionally deferred (see follow-ups).
- **Migration tests** under `finagent/tests/integration/`: ephemeral
  pgvector container per session, one smoke probe per migration, and a
  behavioral RLS probe that creates a `NOSUPERUSER NOBYPASSRLS` role
  and asserts the policy returns the right row count per `SET app.manager_id`.
- **Decoupling**: `alembic/env.py` reads `DATABASE_URL` from
  `os.environ` directly, not via `app.config.settings`. Removes a
  fragile import-order dependency on Pydantic-settings' freezing.

## What this PR does NOT do

- RLS on the remaining five tables (`Manager`, `Company`, `Portfolio`,
  `PortfolioHolding`, `Recommendation`) — tracked in a follow-up issue.
- HNSW parameter tuning (`m`, `ef_construction`) — deferred until we
  have a query corpus to measure recall against.
- `.gitignore` updates (`.venv/`, `.pytest_cache/`, IDE patterns) —
  small follow-up commit; tracked separately.

## How to verify

Requires Docker Desktop running.

```bash
# Expected: 5 passed in ~6-7s against an ephemeral pgvector container.
```

```bash
cd finagent && uv run mypy tests/
# Expected: Success, no issues found.
```

```bash
cd finagent && uv run alembic upgrade head
# Expected: seven migrations apply cleanly against a configured DATABASE_URL.
```

## Decisions worth flagging

- **Migrations don't import `app.config`.** `env.py` reads
  `DATABASE_URL` from `os.environ` directly. Migrations are a deployment
  tool, not an app tool; they should not depend on the app's settings
  module's import-time freezing behavior.
- **`fetch_scalar` returns `Any`, not `object` or per-call casts.**
  The helper runs genuinely-dynamic SQL where the column type varies
  per call. `Any` is the honest annotation.
- **Behavioral RLS test uses a throwaway role inside the test
  container**, not the eventual app role. The test proves the *policy
  mechanism*; the app-role posture is a separate concern (follow-up).
- **`DROP OWNED BY rls_test CASCADE` is the teardown idiom** for any
  role in a test fixture, not per-grant `REVOKE`. Survives future
  grant additions without test changes.


pradyumna-001 (commented):
Peer review (in-chat Socratic loop) starts here. I'll post one architectural question per comment so we can defend each decision in order. — reviewer

pradyumna-001:
**Q1a (architecture):** `Manager.email` is `String(255)`. Why 255 specifically — what's the contract that picked that number? Two or three sentences, no code.

pradyumna-001:
**Q1b (architecture)** — `finagent/app/db/models.py:16`:

```python
email: Mapped[str] = mapped_column(String(255), unique=True, nullable=False)
```

Why `unique=True` at the column level instead of either a separate `UniqueConstraint` or `Index("ix_managers_email", "email", unique=True)`? What does each shape cost and give you, and why did you pick the inline form? Two or three sentences, no code.

finagent/app/db/models.py
```diff
@@ -0,0 +1,70 @@
...
+
+    id: Mapped[int] = mapped_column(primary_key=True)
+    name: Mapped[str] = mapped_column(String(120), nullable=False)
+    email: Mapped[str] = mapped_column(String(255), unique=True, nullable=False)
+    created_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now(), nullable=False)
+
+
+class Company(Base):
+    __tablename__ = "companies"
+
```
pradyumna-001, line 26: **Q1a (architecture)** — the `String(255)` choice on this line.

Why 255 specifically? Walk me through the contract — where the number came from, and what fails if you're wrong.

**Answer:** 255 isn't a measured p99. The contract is RFC 5321, which caps an email address path at 254 octets in practice; 255 is that bound + one byte of slack. Cheap to revise later via Alembic (`ALTER COLUMN ... TYPE`).

pradyumna-001, line 26: **Q1b (architecture)** — the `unique=True` choice on this line.

**Answer:** Inline `unique=True` because the email itself is the business-uniqueness rule (one manager → one email) and it's a single column. It gives you both the constraint and the backing index in one DDL construct. Separate `UniqueConstraint` is reserved for cases where we need a stable explicit name; a unique `Index` is reserved for partial/composite uniqueness we don't yet have. On `morning_note_id` and other FKs in this file, `index=True` (no `unique`), because the column is high-cardinality across the table but rows-per-value isn't 1.

pradyumna-001: RFC 5321 max = 254
plus 1 byte of slack

finagent/app/db/models.py
```diff
@@ -0,0 +1,70 @@
...
+    id: Mapped[int] = mapped_column(primary_key=True)
+    ticker: Mapped[str] = mapped_column(String(12), unique=True, nullable=False)
+    name: Mapped[str] = mapped_column(String(255), nullable=False)
+    sector: Mapped[str | None] = mapped_column(String(120))
+    created_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now(), nullable=False)
+
+
+class Portfolio(Base):
```
pradyumna-001, line 34: **Q2 (architecture)** — file:24 `String(12)` for `ticker`.

B3 tickers are 4–6 chars in practice (e.g. `PETR4`, `VALE3`, `ITSA4`). Why 12? Where does the contract for that bound come from?

finagent/app/db/models.py
pradyumna-001: Inline `unique=True` because the uniqueness is business rule I want the DB to enforce (one manager, one email). It gives me the rule and the index in one construct. On FKs like morning_note_id where every row has a different value but joins/cascades are real, `index=True` alone covers the access path.

finagent/app/db/models.py
```diff
@@ -0,0 +1,70 @@
...
+    id: Mapped[int] = mapped_column(primary_key=True)
+    portfolio_id: Mapped[int] = mapped_column(ForeignKey("portfolios.id"), nullable=False, index=True)
+    company_id: Mapped[int] = mapped_column(ForeignKey("companies.id"), nullable=False, index=True)
+    weight: Mapped[float] = mapped_column(nullable=False)
+    added_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now(), nullable=False)
+
+
+class MorningNote(Base):
+    __tablename__ = "morning_notes"
+
```
pradyumna-001, line 62: Q3 (architecture) — file:45 `weight: Mapped[float]` on `portfolio_holdings`.

Float vs Numeric. Walk me through what math runs on this column and why float is right here vs `Numeric(precision, scale)`.

finagent/app/db/models.py
pradyumna-001: B3 codes are 4-6 chars. Yahoo's US API suppports fractional share tickers (e.g. PETR4F, 7 chars), which we'll see if we ever ingest US equities too. String(12) is the upper bound of that consumer-Yahoo contract, not of B3 itself. Slack is intentional and unchanged.

finagent/app/db/models.py
pradyumna-001: On this column: rendering + per-portfolio sanity sums. Float drift at 1e-9 isn't a constraint violation we need to enforce in the DB, it's a display-time rounding concern that the cart layer already handles. Numeric(precision, scale) will appear on money-shape columns (positions, PnL), where drift could violate an invatiant - that hasn't landed yet.

finagent/app/db/models.py
```diff
@@ -0,0 +1,70 @@
...
+    id: Mapped[int] = mapped_column(primary_key=True)
+    morning_note_id: Mapped[int] = mapped_column(ForeignKey("morning_notes.id"), nullable=False, index=True)
+    action: Mapped[str] = mapped_column(String(16), nullable=False)
+    confidence: Mapped[float] = mapped_column(nullable=False)
+    justification: Mapped[str] = mapped_column(Text, nullable=False)
+    created_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now(), nullable=False)
+    confirmed_at: Mapped[datetime | None] = mapped_column(DateTime(timezone=True), nullable=True)
+
+
```
pradyumna-001: Q4 (architecture) — file:64 `action: Mapped[str] = mapped_column(String(16))`.

Buy/sell/neutral decisions come in three values. Why free-form `String(16)` instead of an `Enum` type or a small lookup table? Walk me through what would have to change for you to switch.

pradyumna-001: Free-form String(16) for v1. Three strings today, but the value space could grow. Native Postgres Enum would freeze-postgres-values into DDL and we want headroom before we commit.

finagent/app/db/models.py
```diff
@@ -0,0 +1,70 @@
...
+    id: Mapped[int] = mapped_column(primary_key=True)
+    name: Mapped[str] = mapped_column(String(120), nullable=False)
+    email: Mapped[str] = mapped_column(String(255), unique=True, nullable=False)
+    created_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now(), nullable=False)
+
+
+class Company(Base):
+    __tablename__ = "companies"
+
+    id: Mapped[int] = mapped_column(primary_key=True)
```
pradyumna-001, line 27: Q5 (architecture) — file:17 uses `server_default=func.now()` (and the same on every `*_at` column).

You're relying on Postgres to fill the timestamp on every `*_at` column. Two questions packed in one:

(a) Why `func.now()` instead of `DateTime(timezone=True)` plus a Python-side `default=datetime.utcnow` on the ORM? Both look the same to a test — what breaks if you switched?

(b) `func.now()` in Postgres returns `timestamp without time zone`. You declared `DateTime(timezone=True)`. Are those compatible, and which one wins at the DB level?

pradyumna-001: (a) The Python default=datetime.utcnow only kicks in when the ORM constructs the INSERT - raw SQL inserts skit it. That's the breakage if we switched.

(b)  func.now() is NOW() which returns timestamp with time zone in Postgres ≥ 14; SQLAlchemy's DateTime(timezone=True) maps to TIMESTAMP WITH TIME ZONE — types align. The collateral divergence is the connection's session timezone: same value reads back at different wall-clock strings across sessions if host TZ leaks into the container. Today we'd notice by comparing a raw psql INSERT result against an ORM round-trip; if they disagree, v1 ships the AT TIME ZONE 'UTC' pins on read.

finagent/app/db/models.py
```diff
@@ -0,0 +1,70 @@
...
+    portfolio_id: Mapped[int] = mapped_column(ForeignKey("portfolios.id"), nullable=False, index=True)
+    generated_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now(), nullable=False)
+    note_text: Mapped[str] = mapped_column(Text, nullable=False)
+    pipeline_run_id: Mapped[str | None] = mapped_column(String(64), nullable=True, index=True)
+
+
+class Recommendation(Base):
+    __tablename__ = "recommendations"
+
+    id: Mapped[int] = mapped_column(primary_key=True)
```
pradyumna-001, line 84: Q6 (architecture) — file:56 `pipeline_run_id: Mapped[str | None] = mapped_column(String(64), nullable=True, index=True)`.

Two questions in one:

(a) Why `String(64)` and not the native UUID type? Walk me through the trade-off against `from sqlalchemy import UUID` / Postgres `uuid` column type.

(b) The column is nullable AND indexed. Nullable-B-tree-indexes in Postgres don't index NULL values, so `WHERE pipeline_run_id IS NULL` does a sequential scan. Is that intentional, or is this index meant for the populated case only?

pradyumna-001: (a) String(64) over native uuid because we don't yet know the format that downstream pipeline runs will produce. We could push UUID hex (32 chars), UUID-with-dashes (36), or a system-tagged format like runner-2026-07-25T1234-abc. 64 covers all three with slack. Going native uuid later means a USING pipeline_run_id::uuid migration which Postgres tolerates but you'd have to validate every existing value parses — better to commit the column shape when we know the producer.

(b) String(64), nullable=True, index=True: the index fast-paths the populated case (WHERE pipeline_run_id = '<id>'). NULL values aren't in the index because Postgres B-tree defaults skip NULLs — that's fine because v1's access pattern is "fetch notes by run id," not "fetch all NULL notes." If v2 ever needs IS NULL lookups, we'd add a partial index or change the access pattern.

pradyumna-001:
Closing review at Q6. Q7 (confirmed_at) deferred — see #3 follow-ups for context. Will revisit when nullable-timestamp columns come up again.

finagent/tests/integration/test_rls.py
```diff
@@ -0,0 +1,39 @@
...
+async def test_rls_filters_by_manager_id(rls_role: str) -> None:
+    conn = await _connect_as_role(rls_role)
+    try:
+        await _set_manager_id(conn, "2")
+        assert await _count_visible(conn) == 2
+
+        await _set_manager_id(conn, "3")
+        assert await _count_visible(conn) == 1
+
+        await _set_manager_id(conn, None)
```
pradyumna-001, line 31: Q1 (architecture) — finagent/tests/integration/test_rls.py:31 — malformed-input probes.

The test probes four cases in sequence: SET '2' → 2 rows, SET '3' → 1 row, RESET → InvalidTextRepresentationError, SET 'banana' → same error. The first two prove the policy filters correctly. The last two prove that *invalid input* fails loud at the cast boundary, not silently.

Walk me through the decision: was the malformed-input coverage added because the same code path covers it cheaply (an opportunistic extra probe while writing the happy-path), or because it's a separate invariant — 'deny loud on any invalid input' — that you wanted to lock down before merging? In other words: if a future PR refactored `_set_manager_id` to swallow cast errors and return a default value, would the test still catch it, and is that the behavior you want to lock in?

pradyumna-001: Malformed-input coverage was added because it's a separate invariant that I wanted to lock down before merging. Malformed-input coverage was added because it's a separate invariant that I wanted to lock down before merging. If _set_manager_id is refactor, the test still catches it. If _set_manager_id is refactor, the test still catches it.

finagent/tests/conftest.py
```diff
@@ -0,0 +1,91 @@
...
+        yield url
+    finally:
+        if original_db_url is None:
+            os.environ.pop("DATABASE_URL", None)
+        else:
+            os.environ["DATABASE_URL"] = original_db_url
+
+
+@pytest.fixture
+def rls_role(migrated_db_url: str) -> Generator[str, None, None]:
```
pradyumna-001, line 42: Q2 (architecture) — finagent/tests/conftest.py:42 — rls_role fixture seeds three morning_notes rows with manager_id 2 having two rows and manager_id 3 having one.

The fixture hardcodes specific row counts to make the policy probes assertable (2 rows for manager 2, 1 for manager 3). Walk me through the decision: why those exact numbers, and why that asymmetry (2 vs 1, not 2 vs 2)? What does this row distribution prove that a symmetric 1-vs-1 setup wouldn't?

pradyumna-001: The assymetry proves the policy keys off the per-row manager_id, not off a constant row count or first-N-rows heuristic. A symetric setup would still pass even if the policy were buggy as "always return 2 rows".

finagent/alembic/versions/1d4e6f8a2b35_enable_rls_on_morning_notes_.py
```diff
@@ -0,0 +1,30 @@
...
+Create Date: 2026-07-28 21:00:00.000000
+
+"""
+from typing import Sequence, Union
+
+from alembic import op
+
+
+revision: str = "1d4e6f8a2b35"
+down_revision: Union[str, Sequence[str], None] = "9c3d5e7f1a23"
```
pradyumna-001, line 14: Q3 (architecture) — finagent/alembic/versions/1d4e6f8a2b35_enable_rls_on_morning_notes_.py:14 — current_setting('app.manager_id', true)::int.

The RLS policy casts `current_setting('app.manager_id', true)` to int. The `true` second-argument means 'missing OK, return NULL.' The cast then turns NULL into an `InvalidTextRepresentationError` — which is what the test asserts (RESET → error).

So the chain is: missing-OK at the setting level → NULL → cast fails → query errors. My read is that the error surface (`InvalidTextRepresentationError`) is a side effect of the type cast, not a designed denial. Is that the right read, or did you intend the missing-setting case to be a first-class 'denied loud' state with its own error type?

pradyumna-001: The denied-loud state surfaces as InvalidTextRepresentationError at the cast boundary; no typed exception needed because the PostgreSQL error is already loud and surfaces immediately in any handler. If a future layer wants a typed MissingManagerIdError, that's app-level exception translation, not a migration concern.

finagent/tests/conftest.py
```diff
@@ -0,0 +1,91 @@
...
+    async def _teardown() -> None:
+        conn = await _admin()
+        try:
+            await conn.execute("DROP OWNED BY rls_test CASCADE")
+            await conn.execute("DROP ROLE IF EXISTS rls_test")
+        finally:
+            await conn.close()
+
+    asyncio.run(_setup())
+    try:
```
pradyumna-001, line 88: Q4 (architecture) — finagent/tests/conftest.py:88 — `yield f"postgresql+asyncpg://rls_test:rls_test_pwd@{admin_url.split('@', 1)[1]}"`.

The fixture rebuilds a connection URL by string-splitting `admin_url` on `@` and re-prefixing the scheme. This is brittle in two ways I can see: (1) if the admin password contains `@`, the split captures the wrong segment; (2) if `admin_url` is `postgresql://` (no driver suffix) the literal substring `postgresql+asyncpg` replacement earlier might leave a malformed URL.

Walk me through why you chose string surgery over constructing the URL from components (e.g. `urllib.parse` or storing admin user/password/host/port separately at fixture setup). What does the string-split version gain?

pradyumna-001: String-split because it's a one-liner and the current password (rls_test_pwd) has no URL-significant chars, so the brittleness is theoretical. Revisit with urllib.parse if the password ever contains @ or if we add more roles."

pradyumna-001:
Closing this PR as part of the project rebuild.

The data-layer design in `rebuild/issue-03-models` (six tables with `portfolio_holdings` as a separate entity, generic `id` PKs, `HALFVEC(2048)` embedding, `CheckConstraint`-based status, `JSONB` non-null flags) diverges from the schema already shipped to `main` (M:M `portfolios` join table, named PKs like `manager_id`/`morning_note_id`, `Vector(1536)`, `Enum`, `ARRAY(JSONB)` flags).

The rebuild posture treats issue #03 as the actual first coding issue. `main` is a historical artifact from the previous attempt — we are not merging into it; we are replacing it. `rebuild/issue-03-models` becomes the new `main`.

The migration tests, RLS behavioral probe, and `fetch_scalar` helper from this branch carry forward; the data-layer shape changes are tracked in follow-up issues.

Issue #34, opened 2026-07-30
#04a — FastAPI lifespan, async engine, connection pool, and /health


pradyumna-001:
## Overview

First chunk of issue #04. Establishes the FastAPI application
skeleton: async engine + connection pool on startup, clean disposal on
shutdown, and a \`/health\` endpoint that pings the database with
\`SELECT 1\`.

This is the smallest possible slice of #04 that produces a runnable
service. No business logic, no RLS-touching endpoints, no logging
middleware — those land in #04b and #04c.

## Scope

1. **\`app/main.py\`** — FastAPI app factory with \`lifespan\` context
   manager that:
   - Creates an async SQLAlchemy engine on startup.
   - Disposes the engine on shutdown (\`engine.dispose()\`).
2. **\`app/db/session.py\`** — async session factory and
   \`get_session\` dependency that yields a session per request.
3. **\`GET /health\`** — returns
   \`{"status": "ok", "db": "ok"}\`. Performs \`SELECT 1\` against the
   engine. Returns 503 if the DB ping fails.
4. **\`app/config.py\` update** — keep \`DATABASE_URL\` (will be split
   in #03b; not addressed here).
5. **Smoke probe** — \`tests/integration/test_health.py\` (or extend
   existing \`tests/integration/\`) that boots the app via
   \`httpx.AsyncClient\` against the ephemeral pgvector container and
   asserts \`/health\` returns 200 with \`db: ok\`.

## Acceptance criteria

- [ ] App boots with \`uv run uvicorn app.main:app\`.
- [ ] \`/health\` returns 200 with \`{"status": "ok", "db": "ok"}\`.
- [ ] \`/health\` returns 503 if \`DATABASE_URL\` is unreachable.
- [ ] Engine disposed cleanly on shutdown (no leaked connections in
  the ephemeral container).
- [ ] \`mypy app/\` clean.
- [ ] \`pytest tests/integration/\` passes.

## Out of scope

- Logging middleware (#04b).
- \`/morning-notes\` and \`gestor-id\` enforcement (#04b).
- \`POST /pipeline/trigger\` (#04c).
- SSE endpoint (#04c).
- App role split (#03b).
- RLS on remaining tables (#03a).

## How to verify

\`\`\`bash
cd finagent && uv run uvicorn app.main:app --port 8000
# In another shell:
curl http://localhost:8000/health
# Expected: {"status":"ok","db":"ok"}

cd finagent && uv run pytest tests/integration/
[...]

Issue #7, opened 2026-07-14
#07: feat: define Typed AgentState schema, model types, and validation engine


pradyumna-001:
## Overview
Implement the strictly typed state management architecture for our multi-agent LangGraph execution flow. This involves creating helper structures (`MacroOutput`, `CompanyEvent`, etc.), defining the core `AgentState` schema, and establishing invariant validation hooks to prevent corrupted states from executing downstream.

> **Architecture Convention Reminder:** Consistent with our English language standards, `gestor_id` will be defined as `manager_id` across the schemas and state dictionaries.

### 1. Model Definitions (`app/graph/state.py`)
Create Pydantic/Typed representation models for the individual agent outputs and pipeline flags:
* **`MacroOutput`**: Structured macro-economic indicators and contextual text.
* **`CompanyEvent`**: Represents key corporate developments, dates, and significance levels.
* **`QuantOutput`**: Standard quantitative metrics, trends, and signals.
* **`RiskFlag`**: Specific risk classifications (market, operational, regulatory) with severity levels.
* **`DataFlag`**: Built-in pipeline warning flags (aligned with our existing Pydantic model structure).
* **`Recommendation`**: Final investment/allocation thesis with targeted weights and target horizons.

### 2. Core Orchestration State (`AgentState`)
Define the workflow state as a comprehensive `TypedDict`:
- **Core Identifiers:**
  - `pipeline_run_id`: `str`
  - `morning_note_id`: `str`
  - `manager_id`: `int` (mapped from `gestor_id`)
  - `empresa_ticker`: `str`
- **Contextual Agent Outputs:**
  - `macro_context`: `MacroOutput | None`
  - `company_events`: `list[CompanyEvent]`
  - `quant_metrics`: `QuantOutput | None`
  - `risk_flags`: `list[RiskFlag]`
- **Synthesized Deliverables:**
  - `morning_note`: `str | None`
  - `recommendation`: `Recommendation | None`
- **Metadata & Guardrails:**
  - `confidence_scores`: `dict[str, float]`
  - `data_freshness`: `dict[str, datetime]`
  - `flags`: `list[DataFlag]`

### 3. State Invariant Guardrails
* **`create_initial_state()`**: Generates a default `AgentState` with empty list/dict structures initialized correctly.
* **`validate_state(state: AgentState)`**: An assertion-based validator run before each node execution to verify system invariants. It must strictly validate:
[...]

PR #37, opened 2026-07-30
feat(issue-04a): FastAPI lifespan, async engine, /health

pradyumna-001:
## What this PR does

- **app/core/config.py** — pydantic-settings `Settings` class with `DATABASE_URL` and `MIGRATION_DATABASE_URL` as required fields; module-level `settings = Settings()` singleton. Env access isolated to one file.
- **app/db/session.py** — `AsyncEngine` + `async_sessionmaker` bound to `settings.DATABASE_URL`; `get_session()` async generator dependency for FastAPI request lifecycle.
- **app/main.py** — `@asynccontextmanager` lifespan that runs `SELECT 1` at startup (fail-fast if DB unreachable, uvicorn refuses to bind) and `await engine.dispose()` at shutdown. `GET /health` returns `{"status":"ok","db":"ok"}` on `SELECT 1` success, `503` with `{"status":"degraded","db":"unreachable"}` on `SQLAlchemyError`.
- **tests/integration/test_health.py** — `httpx.ASGITransport` test that monkeypatches `app.main.engine` to raise `OperationalError`, asserts 503 contract. No live DB needed.
- **.env** — added `MIGRATION_DATABASE_URL` sibling to `DATABASE_URL` (psycopg sync DSN per #03b design).
- **pyproject.toml** — added `httpx` to `[dependency-groups].dev`; added `[[tool.mypy.overrides]]` for `app.core.config` (mypy cant see `.env` at type-check time, so the env-binding module is exempted).

## What this PR does NOT do

- `/morning-notes` endpoint and `gestor-id` header enforcement — #04b.
- `POST /pipeline/trigger` and SSE — #04c.
- `MIGRATION_DATABASE_URL` plumbing into Alembic `env.py` — #03b.
- Logging middleware, structured logging, `ContextVars` — #04b.

## How to verify

```bash
cd finagent
uv run pytest tests/                          # 6 passed
uv run mypy tests/ app/                       # 0 errors
uv run ruff check .                           # 0 errors

uv run docker compose up -d postgres_vector
uv run uvicorn app.main:app --port 8000       # lifespan probe runs SELECT 1
curl http://localhost:8000/health             # 200 + {"status":"ok","db":"ok"}
```

The 503 path is covered by the unit test (no live DB required):

```bash
uv run pytest tests/integration/test_health.py -v   # test_health_returns_503_when_db_raises PASSED
```

To exercise the lifespan rejection path manually:

```bash
uv run docker compose stop postgres_vector
uv run uvicorn app.main:app --port 8000       # uvicorn exits: Application startup failed
```

## Decisions worth flagging

1. **`settings = Settings()` at module level in `config.py`, not inside `session.py`.** The env-to-Python boundary lives in one module. `session.py` consumes; nothing else reads env directly. This is the seam that lets `#03b` split app vs. migration URLs cleanly later.

2. **`SQLAlchemyError` catch in `/health`, not `Exception`.** A 503 must mean "DB problem," not "I have a bug." `Exception` would mask programmer errors (TypeError, AttributeError) as DB outages, which corrupts alerting.

3. **Lifespan probe is fail-fast (no serve on DB unreachable at boot).** Per the failure-handling semantic, a process that cant reach its DB shouldnt bind the port and serve 503s forever — it should crash and let the orchestrator restart or replace it. This makes the `SELECT 1` at startup the loud boundary, not `/health`.

4. **`engine.begin()` used in both lifespan and `/health` (transactional).** Overkill for `SELECT 1` today, but the right primitive if we ever extend the lifespan probe to run `CREATE EXTENSION` or schema checks. Pick the safer one now; cheaper than a refactor later.

5. **Pre-existing cruft deleted.** `app/config.py`, `app/graph/state.py`, `tests/test_config.py` were orphans from the historical `main` that the rebuild posture (see journal #07) was supposed to clear. They had no importers. `ruff --fix` cleared 5 more unused imports in `app/api/deps.py`, `app/db/models.py`, `tests/conftest.py`. None of these are part of #04as feature scope, but all caused red CI/mypy/ruff today.

6. **Test uses `ASGITransport`, not `uvicorn`-in-process.** Lifespan doesnt run under ASGITransport, which is correct for this test: we want to control the failure path explicitly. Future integration tests that need a real DB should use `testcontainers[postgres]` (already in deps) and a session-scoped fixture.

Refs #34


finagent/app/main.py
```diff
@@ -1,11 +1,27 @@
...
-from app.core.config import settings
 
-app = FastAPI (
-    title=settings.app_name,
-    version="0.1.0"
-)
+@asynccontextmanager
+async def lifespan(app: FastAPI):
+    async with engine.begin() as conn:
+        await conn.execute(text("SELECT 1"))
```
pradyumna-001, line 12: Q1 (architecture) — app/main.py:18  lifespan: SELECT 1 fail-fast.

You said the orchestrator should restart or replace a process that can't reach its DB. Walk me through that contract. When uvicorn exits because the lifespan probe fails, what restarts it — ECS task definition, k8s livenessProbe, plain systemd? And what does the restart *prove*? (A 30-second restart loop on a permanently unreachable DB is a different problem than a 5-minute crashloop backoff.)

finagent/tests/integration/test_health.py
```diff
@@ -0,0 +1,28 @@
+import pytest
+from httpx import ASGITransport, AsyncClient
+from sqlalchemy.exc import OperationalError
+from app.main import app
+
+
+@pytest.mark.asyncio
+async def test_health_returns_503_when_db_raises(monkeypatch: pytest.MonkeyPatch) -> None:
```
pradyumna-001, line 8: Q4 (architecture) — tests/integration/test_health.py:8  ASGITransport, no lifespan.

The test stubs  to raise OperationalError — but lifespan *also* calls  at startup. Under ASGITransport lifespan doesn't run, so the test gets a clean 503 from /health without ever hitting the lifespan path. Two questions: (a) does anything test the *real* lifespan SELECT 1 path against the real testcontainer DB today? (b) if not, the lifespan probe is shipped untested in production — is that acceptable, and why?

finagent/app/db/session.py
```diff
@@ -0,0 +1,15 @@
+from collections.abc import AsyncGenerator
+from sqlalchemy.ext.asyncio import (
+    AsyncSession,
+    async_sessionmaker,
+    create_async_engine,
+)
+from app.core.config import settings
```
pradyumna-001, line 7: Q3 (architecture) — app/db/session.py:7  asyncpg driver choice.

You picked asyncpg over psycopg3 async. Both are stable in 2026. What did asyncpg give you that psycopg3 async didn't, and what's the cost you accepted by picking it? (Specific not vibes — e.g., JSONB roundtrip performance, prepared-statement cache behavior, ecosystem maturity for sqlalchemy 2.x async.)

finagent/app/main.py
```diff
@@ -1,11 +1,27 @@
...
+    yield
+    await engine.dispose()
 
-app.include_router(router)
+
```
pradyumna-001, line 16: Q2 (architecture) — app/main.py:23  /health catches SQLAlchemyError, not Exception.

Defense in the PR body: '503 must mean DB problem, not I have a bug.' Concretely: if a request raises an AttributeError (e.g., settings misconfigured), what does the user see and what does CloudWatch see? Walk me through the chain — status code, response body, log line, alert. The shape of that chain is what makes the catch narrow enough.

Issue #35, opened 2026-07-30
#04b — Logging middleware with correlation IDs and GET /morning-notes


pradyumna-001:
## Overview

Second chunk of issue #04. Adds the structured-logging middleware
that threads \`pipeline_run_id\` and \`morning_note_id\` through every
log record via \`contextvars\`, plus the first RLS-touching endpoint:
\`GET /morning-notes\` filtered by the \`gestor-id\` header.

The middleware contract here is what the agents (issues #08–13) and
the pipeline (issue #14) will inherit. Get the shape right now and
the rest of the system inherits clean observability.

## Scope

1. **\`app/core/logging_config.py\`** — structured logger configured
   with a filter that reads \`pipeline_run_id\` and
   \`morning_note_id\` from \`contextvars\` and appends them to every
   record.
2. **ContextVar definitions** — \`current_pipeline_run_id\`,
   \`current_morning_note_id\` (or one combined \`correlation_ids\`
   TypedDict).
3. **ASGI middleware** — \`app/core/middleware.py\`:
   - Reads \`X-Pipeline-Run-Id\` and \`X-Morning-Note-Id\` headers if
     present, sets the contextvars for the duration of the request,
     resets on exit.
   - Logs request start/end with method, path, status, duration.
4. **\`GET /morning-notes\`** — returns notes for the current
   \`gestor-id\`. Requires \`gestor-id\` header; returns 400 if
   missing. Sets \`app.manager_id\` GUC via \`SET LOCAL\` so RLS
   filters correctly.
5. **Smoke probes** — extend \`tests/integration/\`:
   - Middleware injects and clears contextvars (unit test).
   - \`GET /morning-notes\` returns 400 without header.
   - \`GET /morning-notes\` returns 200 with header, filtered by
     RLS.

## Acceptance criteria

- [ ] Every log line during a request that sets
  \`X-Pipeline-Run-Id\` includes that ID in the structured output.
- [ ] ContextVars cleared on request exit (no leak between requests).
- [ ] \`GET /morning-notes\` returns 400 without \`gestor-id\` header.
- [ ] \`GET /morning-notes\` returns only the notes for the
  request's \`gestor-id\`, enforced by RLS (the behavioral RLS probe
  confirms this; an end-to-end probe confirms it via the HTTP layer).
- [ ] \`mypy app/\` clean.
- [ ] \`pytest tests/integration/\` passes.

## Out of scope

- Pipeline trigger and SSE (#04c).
[...]

Issue #24, opened 2026-07-19
Implement EditorAgent Node


pradyumna-001:
## Description
Implement the EditorAgent node inside `app/agents/editor.py` as the final consolidation layer of the multi-agent pipeline. This agent ingests all upstream analytical structures (`macro_context`, `company_events`, `quant_metrics`, and `risk_flags`), reconciles their content, populates structured section-by-section confidence scores, and compiles a comprehensive financial morning note written in Portuguese alongside a structured, typed recommendation.

## Tasks
- Create the file `app/agents/editor.py`.
- Implement the entrypoint function `editor_agent_node(state: AgentState) -> AgentState` with transaction context tracking logging.
- Configure the agent execution block to use `gpt-4o` for synthesizing complex investment theses and professional editorial layout generation.
- Enforce language requirements: the generated text for `state['morning_note']` must be fully written in Portuguese.
- Format the morning note with explicit, mandatory markdown sections: Macro Context, Corporate Events, Quantitative Metrics, Risk Factors, and Investment Recommendation.
- Implement the structured `confidence_scores` mapping logic across sections using upstream metrics:
  - If any section/upstream node generated a `DataFlag` warning or experienced an omission, penalize that specific section's confidence score below `0.5` and force an explicit, visible warning notice directly inside that section's text body (Fail Visible paradigm).
- Parse the structured final investment recommendation and populate `state['recommendation']` with a typed Pydantic `Recommendation` model containing `action` (BUY, HOLD, SELL, UNDERWEIGHT), `target_weight`, `horizon_months`, and `thesis_summary`.
- Update `state['data_freshness']['editor']` with the current ISO timestamp upon completion.
- Create an isolated unit test suite verifying that the EditorAgent creates a readable morning note markdown string and structural recommendation fields under passing parameters.

### 1. Fail Visible Mandate
If data omissions occur up the chain (e.g., your `macro_context` missing error caught in Task 12), the EditorAgent must not hide the issue. It must display an uppercase hazard banner within that section to make the pipeline failure transparent to end portfolio managers.

### 2. Multi-Class Schema Integrity
The node must output dual parameters: a raw formatted markdown text string into `state['morning_note']` and a completely instantiated `Recommendation` Pydantic model into `state['recommendation']`.

---

## Expected Deliverables
- A standalone, fully typed final assembly node at `app/agents/editor.py`.
[...]

Issue #9, opened 2026-07-14
feat: implement MacroAgent node with Tavily Search and GPT-4o-mini integration 


pradyumna-001:
## Description
Implement the `MacroAgent` node responsible for gathering, analyzing, and synthesizing Brazil's macroeconomic context using Tavily Search and GPT-4o-mini. The node must strictly populate the `macro_context` in our `AgentState` or gracefully append a safety flag if external integrations fail.

## Tasks
- [ ] Create `app/agents/macro.py`
- [ ] Implement `macro_agent_node(state: AgentState) -> AgentState`
- [ ] Integrate **Tavily API** to search for Brazilian macroeconomic updates (Focus sources: *Banco Central, IBGE, Reuters BR, Bloomberg Brasil*)
- [ ] Use **GPT-4o-mini** to extract, analyze, and synthesize the search results into structured data
- [ ] Populate `state["macro_context"]` using the typed `MacroOutput` Pydantic model
- [ ] Record metadata inside `state["data_freshness"]["macro"]` with the current timestamp
- [ ] Handle failures gracefully: If Tavily or the LLM call fails, append a structured `DataFlag` to `state["flags"]` and return the state with `macro_context=None` (no silent failures!)
- [ ] Ensure comprehensive structured logging using `pipeline_run_id` and `morning_note_id` for traceability

## Definition of Done
- A fully functional `macro_agent_node` that processes live macroeconomic data.
- Unit tests written under `tests/unit/test_macro.py` validating both the happy path (successful API results) and the fallback path (Tavily failure and flag appending).
- Test suite runs and passes cleanly with `uv run pytest`.

PR #38, opened 2026-07-30
feat(issue-04b): logging middleware, GET /morning-notes, RLS context

pradyumna-001:
## What this PR does

- **app/core/context.py** — module-level `ContextVar`s for `current_pipeline_run_id` and `current_morning_note_id`, default `None`. Producer for context, consumed by logger and middleware.
- **app/core/logging_config.py** — `ContextFilter` enriches every `LogRecord` with the ContextVar values; `configure_logging()` installs a `StreamHandler` with a format string that reads them.
- **app/middleware/correlation.py** — class-based `BaseHTTPMiddleware` that reads `X-Pipeline-Run-Id` and `X-Morning-Note-Id` headers, sets ContextVars for the request lifetime, **resets via `finally`** on exit (mandatory — failure to reset leaks IDs across requests).
- **app/api/routes/morning_notes.py** — `GET /morning-notes` requires `manager-id` header (`HTTPException(400)` if missing); runs `SET LOCAL app.manager_id = '<int>'` inside the transaction (f-string, not bind params — `SET LOCAL` is a utility command, not a prepared statement); explicit `WHERE manager_id = :mid` defense-in-depth against RLS being disabled.
- **app/main.py** — `app.add_middleware(CorrelationMiddleware)` + `app.include_router(morning_notes_router)`.
- **tests/unit/test_correlation.py** — two probes: ContextVars reset after request with headers, ContextVars stay `None` after request without headers. Catches the reset-skip bug.
- **tests/integration/test_morning_notes.py** — two probes: 400 path returns `{"detail": "manager_id header required"}`, 200 path returns a list.

## What this PR does NOT do

- SSE / `POST /pipeline/trigger` — #04c.
- `MIGRATION_DATABASE_URL` split into the role-based `#03b` work proper (just env.py refactor here; role split + `finagent_app` user is still a separate issue).
- LangSmith / CloudWatch integration — #23, #24.
- Tests for actual RLS row filtering (covered by `test_rls.py`).
- Translating `docs/issues.md` from Portuguese to English (deferred per session discussion).

## How to verify

```bash
cd finagent
uv run pytest tests/                          # 10 passed (4 new + 6 existing)
uv run mypy app/ tests/                       # 0 errors
uv run ruff check .                           # 0 errors

uv run docker compose up -d postgres_vector
uv run alembic upgrade head                   # uses MIGRATION_DATABASE_URL now
uv run uvicorn app.main:app --port 8000

curl -s -w "\nHTTP %{http_code}\n" http://localhost:8000/morning-notes
# {"detail":"manager_id header required"}    HTTP 400

curl -s -w "\nHTTP %{http_code}\n" \
  -H "manager-id: 1" http://localhost:8000/morning-notes
# []                                          HTTP 200
```

Manual ContextVar verification (run a request, then assert the IDs are gone):

```bash
uv run python -c "
import asyncio
from httpx import ASGITransport, AsyncClient
from app.core.context import current_pipeline_run_id
from app.main import app

async def main():
    transport = ASGITransport(app=app)
    async with AsyncClient(transport=transport, base_url='http://test') as c:
        await c.get('/morning-notes', headers={'X-Pipeline-Run-Id': 'pipe-1'})
    print('after request:', current_pipeline_run_id.get())

asyncio.run(main())
"
# after request: None
```

## Decisions worth flagging

1. **HTTP header name is `manager-id`, not `gestor-id`.** The issues doc specifies `gestor-id`; user decision in this session was to use English wire-format throughout the codebase and translate the spec later. PR reviewer will see the spec mismatch — flagging it here so it's explicit. When the spec is translated, the `alias` and `detail` strings in `morning_notes.py` change in lockstep.

2. **`SET LOCAL app.manager_id = '<int>'` uses an f-string, not bind params.** PostgreSQL rejects `SET LOCAL ... = $1` because `SET` is a utility command routed through the simple query protocol, not the extended query protocol (no prepared statements). Defense-in-depth: the value is `int` (Pydantic-validated `Annotated[int | None, Header(...)]`), then `int(...)` cast before f-string interpolation. If anyone changes the type annotation to `str`, this becomes an injection vector — the PR description documents this so the next reader knows.

3. **`alembic/env.py` refactor was originally #03b's scope, accelerated for #04b verification.** The route couldn't work against an unmigrated database, and the migration command was broken (`os.environ["DATABASE_URL"]` hardcoded, no psycopg). The refactor: `Settings.MIGRATION_DATABASE_URL` as the source of truth, with `os.environ["MIGRATION_DATABASE_URL"]` as an override (test fixture uses this). Engine switched from async to sync (psycopg). This is the env.py decoupling PR #30 promised; #03b's remaining work (split DDL/DML grants, `finagent_app` role) is still a separate issue.

4. **`psycopg[binary]>=3.1` added to dependencies.** It was missing despite `MIGRATION_DATABASE_URL` pointing at psycopg. Discovered when alembic failed with `ModuleNotFoundError: No module named 'psycopg'`. The dependency should have been in pyproject.toml when the URL was first introduced in #04a's `.env` edit; missed then, fixed now.

5. **`WHERE manager_id = :mid` is redundant with RLS but kept.** If RLS is ever disabled (misconfigured role, admin debug session), the WHERE still filters. Costs a few microseconds per query; protects against an entire class of bug. The `SET LOCAL` ensures RLS sees the right manager_id even when the WHERE is removed.

6. **Correlation middleware uses `BaseHTTPMiddleware` (class-based), not `@app.middleware("http")` (decorator).** Class-based gives a stable import path and explicit ordering via `add_middleware()`. Decorator order depends on file layout — fragile.

Refs #35


finagent/app/middleware/correlation.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class CorrelationMiddleware(BaseHTTPMiddleware):
+    async def dispatch(self, request: Request, call_next) -> Response:
+        pipeline = request.headers.get("X-Pipeline-Run_Id")
+        note = request.headers.get("X-Morning-Note-Id")
+        p_token = current_pipeline_run_id.set(pipeline) if pipeline else None
+        n_token = current_morning_note_id.set(note) if note else None
+        
+        try:
```
pradyumna-001, line 14: Q5 (architecture) — finagent/app/middleware/correlation.py:14  reset in finally, not except.

The PR body calls this 'the only thing preventing correlation-ID leak across requests on the same uvicorn worker.' Two questions: (a) what does 'leak' actually look like in production logs if the reset is removed — does a future request's logs carry the previous request's pipeline_run_id, and how would an operator detect this from log search alone? (b) test_correlation.py asserts the ContextVar is None after the request — but it doesn't test what happens during the request. Is there a probe that asserts the ContextVar *is* set during dispatch, or are we only asserting the negative space?

finagent/app/api/routes/morning_notes.py
```diff
@@ -0,0 +1,30 @@
...
+    session: Annotated[AsyncSession, Depends(get_session)],
+    manager_id: int | None = Header(None, alias="manager-id"),
+) -> list[dict]:
+    if manager_id is None:
+        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail="manager_id header required")
+
+    async with session.begin():
+        await session.execute(
+            text(f"SET LOCAL app.manager_id = '{int(manager_id)}'")
+        )
```
pradyumna-001, line 22: Q2 (architecture) — finagent/app/api/routes/morning_notes.py:22  WHERE manager_id = :mid belt-and-suspenders.

If RLS is the security layer, what's the threat model where this WHERE matters? RLS is enabled, app.manager_id is set via SET LOCAL, and the row has manager_id=2 — query without WHERE returns rows for manager 1, 2, 3... and RLS filters to manager 2's rows. WHERE manager_id = :mid gets the same answer. So when does the WHERE actually save you? Two cases I can think of: (a) RLS is disabled for this connection (superuser, BYPASSRLS role), (b) SET LOCAL didn't take effect for some reason (connection pool returning a session with stale state). Which one is the real risk you're defending against, and is there a third?

finagent/alembic/env.py
```diff
@@ -1,90 +1,55 @@
-import asyncio
 from logging.config import fileConfig
 import os
 
 from sqlalchemy import pool
 from sqlalchemy.engine import Connection
-from sqlalchemy.ext.asyncio import async_engine_from_config
+from sqlalchemy.engine import create_engine
```
pradyumna-001, line 7: Q4 (architecture) — finagent/alembic/env.py:9  env.py refactor in scope for #04b.

The PR body says this was accelerated from #03b because the route couldn't be verified without it. Two concerns: (a) does this PR touch any file outside the #04b scope in a way that affects #03b's remaining work — namely the role split and finagent_app user? If so, the boundary between what's done here and what's left for #03b isn't crisp. (b) What does a reviewer of #03b learn when they read its open PR vs. when they find half of it in this PR's history? Walk me through the order-of-operations from a future maintainer's view.

finagent/app/api/routes/morning_notes.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+@router.get("")
+async def list_morning_notes(
+    session: Annotated[AsyncSession, Depends(get_session)],
+    manager_id: int | None = Header(None, alias="manager-id"),
+) -> list[dict]:
+    if manager_id is None:
+        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail="manager_id header required")
+
```
pradyumna-001, line 18: Q1 (architecture) — finagent/app/api/routes/morning_notes.py:18  SET LOCAL via f-string, not bind params.

The PR body names this as a foot-gun: 'if anyone changes the type annotation to str, this becomes injection.' Two layers today — Pydantic int validation and an int() cast. Walk me through the failure mode precisely: a future maintainer removes the int() cast (silent refactor), the type stays Annotated[int | None, ...], and FastAPI passes the header through Header(None, alias='manager-id'). What does Pydantic actually do at the boundary, and where does the int() cast bite? I want to know whether the two layers are *both* active at the same time, or whether one is redundant defense that's not actually checking anything.

finagent/app/api/routes/morning_notes.py
```diff
@@ -0,0 +1,30 @@
...
+from fastapi import APIRouter, Depends, Header, HTTPException, status
+from sqlalchemy import text
+from sqlalchemy.ext.asyncio import AsyncSession
+from app.db.session import get_session
+
+
+router = APIRouter(prefix="/morning-notes", tags=["morning-notes"])
+
+
+@router.get("")
```
pradyumna-001, line 11: Q3 (architecture) — finagent/app/api/routes/morning_notes.py:11  manager-id header (English, not spec's gestor-id).

The PR body flags this as a wire-format-vs-spec mismatch with translation deferred. The user decision in the session was 'English everywhere in the codebase.' Three concrete questions: (a) what does this decision do to the issues doc — is it a docs change that needs to land alongside, or does the doc stay Portuguese and we just have a translation debt? (b) clients calling GET /morning-notes will send whatever header name the spec says; if they send gestor-id, do they get 400 or do we accept both? (c) when the spec gets translated, what's the lockstep update you mentioned — same PR or separate?

pradyumna-001:
Heads up for future readers: review of this PR happened on the merged-but-then-reverted tree (see main history: 8208dec was the #04b merge, 70d71a3 reverted it, 68bd60d brought #04a in). Decisions on this PR were discussed in chat during session 2026-07-31; defenses are deferred. The PR #37 review comments were a precedent for this pattern. PR #38 will be re-merged after the reviewer (user) answers the five anchored questions and clears any follow-up issues.

pradyumna-001:
Update 2026-07-31: this PR is effectively superseded by the flatten+redo decision. The 5 architectural questions answered here remain unanswered but the implementation will be redone as part of #04c on top of the flat tree.

Refs journal #09 for the full session record.

PR #39, opened 2026-08-03
feat(issue-04b): logging middleware, GET /morning-notes, RLS context

pradyumna-001:
## What this PR does

Restores #04b implementation to root paths on the post-flatten tree.

- app/core/context.py — ContextVars current_pipeline_run_id, current_morning_note_id
- app/core/logging_config.py — stdlib logging with ContextFilter that reads contextvars
- app/middleware/correlation.py — ASGI middleware reading X-Pipeline-Run-Id / X-Morning-Note-Id headers, resetting contextvars on exit
- app/api/routes/morning_notes.py — GET /morning-notes: 400 without manager-id header, SET LOCAL app.manager_id + filtered query
- app/main.py — wires middleware + router
- tests/integration/test_morning_notes.py — endpoint probes (400 missing-header, 200 with header)
- tests/unit/test_correlation.py — middleware probes (headers set/reset contextvars)

## What this PR does NOT do

- alembic/env.py refactor from original #04b — references MIGRATION_DATABASE_URL, which is #03b contract (unmerged). Forward-ref deferred.
- tests/conftest.py change from original #04b — same reason; sets MIGRATION_DATABASE_URL env var that nothing reads yet.
- manager-id/gestor-id bilingual drift — kept manager-id on the wire (matches SQLAlchemy column manager_id). Translation of docs is a docs follow-up, not a wire change.

Refs #35

## How to verify

```bash
uv run pytest tests/ -v
uv run mypy app/ tests/
uv run ruff check .
```

Expected: 10 tests pass (was 6 baseline + 4 new), mypy clean, ruff clean.

## Decisions worth flagging

- **bound_engine fixture in tests/integration/test_morning_notes.py.** The singleton engine in app/db/session.py is created at module import from settings.DATABASE_URL. The migrated_db_url testcontainer fixture sets DATABASE_URL after import, too late. The fixture monkeypatches app.db.session.engine and SessionLocal to point at the testcontainer URL for the test lifetime. Same pattern as test_health.py 503 stub. Not a reload because reload is fragile across the whole app.
- **Skipped alembic/env.py and conftest.py refactors from original #04b.** Both depend on MIGRATION_DATABASE_URL existing in settings, which is #03b contract. Restoring them now would create a forward dependency on an unmerged issue — bisect break + merge-order ambiguity. Refactors must be self-contained; half-applied refactors are not.
- **Wire header manager-id.** Convention is English on the wire; DB column is manager_id. Portuguese (gestor-id) is a docs translation, deferred.

### DIFF
--- a/app/api/routes/morning_notes.py
+++ b/app/api/routes/morning_notes.py
@@ -1,0 +1,30 @@
+from typing import Annotated
+from fastapi import APIRouter, Depends, Header, HTTPException, status
+from sqlalchemy import text
+from sqlalchemy.ext.asyncio import AsyncSession
+from app.db.session import get_session
+
+
+router = APIRouter(prefix="/morning-notes", tags=["morning-notes"])
+
+
+@router.get("")
+async def list_morning_notes(
+    session: Annotated[AsyncSession, Depends(get_session)],
+    manager_id: int | None = Header(None, alias="manager-id"),
+) -> list[dict]:
+    if manager_id is None:
+        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail="manager_id header required")
+
+    async with session.begin():
+        await session.execute(
+            text(f"SET LOCAL app.manager_id = '{int(manager_id)}'")
+        )
+        result = await session.execute(
+            text(
+                "SELECT id, generated_at, note_text, status FROM morning_notes "
+                "WHERE manager_id = :mid ORDER BY generated_at DESC LIMIT 50"
+            ), {"mid": manager_id},
+        )
+
+        return [dict(row._mapping) for row in result]
--- a/app/main.py
+++ b/app/main.py
@@ -4,6 +4,8 @@ from fastapi.responses import JSONResponse
 from sqlalchemy import text
 from sqlalchemy.exc import SQLAlchemyError
 from app.db.session import engine
+from app.middleware.correlation import CorrelationMiddleware
+from app.api.routes.morning_notes import router as morning_notes_router
 
 
 @asynccontextmanager
@@ -15,6 +17,10 @@ async def lifespan(app: FastAPI):
 
 
 app = FastAPI(title="FinAgent", lifespan=lifespan)
+
+
+app.add_middleware(CorrelationMiddleware)
+app.include_router(morning_notes_router)
 
 
 @app.get("/health")
--- a/tests/integration/test_morning_notes.py
+++ b/tests/integration/test_morning_notes.py
@@ -1,0 +1,36 @@
+from collections.abc import Iterator
+import pytest
+from httpx import ASGITransport, AsyncClient
+from sqlalchemy.ext.asyncio import async_sessionmaker, create_async_engine
+
+import app.db.session as db_session
+from app.main import app
+
+
+@pytest.fixture
+def bound_engine(migrated_db_url: str) -> Iterator[None]:
+    engine = create_async_engine(migrated_db_url, echo=False)
+    db_session.engine = engine
+    db_session.SessionLocal = async_sessionmaker(engine, expire_on_commit=False)
+    yield
+    engine.sync_engine.dispose()
+
+
+@pytest.mark.asyncio
+async def test_list_morning_notes_missing_header_returns_400() -> None:
+    transport = ASGITransport(app=app)
+    async with AsyncClient(transport=transport, base_url="http://test") as client:
+        resp = await client.get("/morning-notes")
+    assert resp.status_code == 400
+    assert resp.json() == {"detail": "manager_id header required"}
+
+
+@pytest.mark.asyncio
+async def test_list_morning_notes_with_header_returns_list(
+    bound_engine: None,
+) -> None:
+    transport = ASGITransport(app=app)
+    async with AsyncClient(transport=transport, base_url="http://test") as client:
+        resp = await client.get("/morning-notes", headers={"manager-id": "2"})
+    assert resp.status_code == 200
+    assert isinstance(resp.json(), list)
--- a/app/core/context.py
+++ b/app/core/context.py
@@ -1,0 +1,5 @@
+from contextvars import ContextVar
+
+
+current_pipeline_run_id: ContextVar[str | None] = ContextVar("current_pipeline_run_id", default=None)
+current_morning_note_id: ContextVar[str | None] = ContextVar("current_morning_note_id", default=None)
--- a/app/core/logging_config.py
+++ b/app/core/logging_config.py
@@ -1,0 +1,25 @@
+import logging
+from app.core.context import current_morning_note_id, current_pipeline_run_id
+
+
+class ContextFilter(logging.Filter):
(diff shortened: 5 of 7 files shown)


app/api/routes/morning_notes.py
```diff
@@ -0,0 +1,30 @@
...
+
+@router.get("")
+async def list_morning_notes(
+    session: Annotated[AsyncSession, Depends(get_session)],
+    manager_id: int | None = Header(None, alias="manager-id"),
+) -> list[dict]:
+    if manager_id is None:
+        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail="manager_id header required")
+
+    async with session.begin():
```
pradyumna-001, line 19: Q1 (architecture) — app/api/routes/morning_notes.py:21 SET LOCAL f-string + int() cast. Two defense layers: int() coerces to integer (kills any non-numeric payload) and single-quote f-string interpolation. If the int() succeeds, the value is safe to interpolate. Walk me through which threat each layer covers and what happens if one is removed.

app/api/routes/morning_notes.py
```diff
@@ -0,0 +1,30 @@
...
+        await session.execute(
+            text(f"SET LOCAL app.manager_id = '{int(manager_id)}'")
```
pradyumna-001, line 21: Q2 (architecture) — app/api/routes/morning_notes.py:23-27 WHERE manager_id = :mid belt-and-suspenders after SET LOCAL. The session is bound to manager_id via app.manager_id GUC, and the WHERE clause also filters by manager_id. If RLS is active, the WHERE is redundant. If RLS fails open (someone disables the policy, role escalation), the WHERE is the only thing keeping manager A from seeing manager B notes. Threat model: when does RLS fail open in this codebase, and is the WHERE enough?

app/api/routes/morning_notes.py
```diff
@@ -0,0 +1,30 @@
...
+from sqlalchemy import text
+from sqlalchemy.ext.asyncio import AsyncSession
+from app.db.session import get_session
+
+
+router = APIRouter(prefix="/morning-notes", tags=["morning-notes"])
+
+
+@router.get("")
+async def list_morning_notes(
```
pradyumna-001, line 12: Q3 (naming) — app/api/routes/morning_notes.py:14 Header alias manager-id vs the SQLAlchemy column manager_id and the docs term gestor-id. Three names for the same concept. PR body says wire matches DB column on purpose and translation is docs follow-up. Walk me through why this asymmetry is correct on the wire even though it drifts from the docs.

tests/integration/test_morning_notes.py
```diff
@@ -0,0 +1,36 @@
...
+import pytest
+from httpx import ASGITransport, AsyncClient
+from sqlalchemy.ext.asyncio import async_sessionmaker, create_async_engine
+
+import app.db.session as db_session
+from app.main import app
+
+
+@pytest.fixture
+def bound_engine(migrated_db_url: str) -> Iterator[None]:
```
pradyumna-001, line 11: Q4 (architecture) — tests/integration/test_morning_notes.py:11-17 bound_engine fixture monkeypatches app.db.session.engine and SessionLocal on each test. Why not importlib.reload? Reload would re-execute the module body with the testcontainer URL set, but reload is brittle — module-level singletons in other modules that have already imported session.engine keep stale references. Walk me through what specifically breaks with reload that monkeypatch handles cleanly.

app/main.py
```diff
@@ -4,6 +4,8 @@
 from sqlalchemy import text
 from sqlalchemy.exc import SQLAlchemyError
 from app.db.session import engine
+from app.middleware.correlation import CorrelationMiddleware
+from app.api.routes.morning_notes import router as morning_notes_router
 
 
 @asynccontextmanager
```
pradyumna-001, line 11: Q5 (scope) — PR body says we skipped alembic/env.py and tests/conftest.py refactors from the original #04b because MIGRATION_DATABASE_URL is #03b's contract. Original #04b had both changes in scope and merged once before. Why is restoring only part of the original scope correct here, and what's the principle that says where to draw the line?

app/api/routes/morning_notes.py
pradyumna-001: the main reason is the docs should be in english, i just didn't translate them yet

app/api/routes/morning_notes.py
pradyumna-001: Defense layer 1: This is a type coercion. If manager_id is "2", int() returns 2. If its a injection, it never executes because the bad value never makes it to SQL. So, it covers non-integer payloads. Strings, SQL fragments, anything that isn't a valid integer literal.

Defense layer 2: Since the first layer garantees we get a int value, the quotes are there to delimit the value as a string. Without them, the SQL command would still work, but if layer 1 ever fails, the quotes are the last line of defense - they ensure that even a non-integer payload becomes a single string literal rather than SQL.

app/api/routes/morning_notes.py
pradyumna-001: When RLS is active, it's redundant. Both layers say the same thing.
Altough, if someone grants BYPASSRLS to the connection role, or the policy is disabled, or GUC is not set, then the WHERE clause is the last line. Without it, RLS-fails-open means every manager's note are visible to anyone. With it, even if RLS is disabled, the application code refuses to return rows that don't match the header-declared manager_id.

tests/integration/test_morning_notes.py
pradyumna-001: Reload is fragile because it leaves stale references in already-imported modules. app.main.py does from app.db.session import engine at import time, which freezes app.main.engine to the old engine object. Reloading app.db.session creates a new engine but doesn't update app.main.engine - it still points to the old one. To fix that you'd need to reload every module that imported from session, transitively.

Monkeypatch works because the codebase accesses the engine through app.db.session.engine and app.db.session.SessionLocal at call time, not through captured local references. Setting db_session.engine = new_engine updates every consumer automatically - app.main.engine, get_session(), anything else that resolver the attribute at access time.

The one assumption that makes monkeypatch safe is no module captures engine into local at module level (e.g., MY_ENGINE = engine). None of our code does that, so monkeypatch is correct.

app/main.py
pradyumna-001: Restoring alembic/env.py to read MIGRATION_DATABASE_URL would break alembic upgrade head because that env var doesn't exist in settings yet — env.py would raise KeyError. Our migrated_db_url fixture runs command.upgrade(cfg, "head") inside it, so the test suite itself would break.

The principle: a PR's scope must be self-contained. Every change must be runnable, testable, and bisectable on its own. Forward-references to unmerged contracts break that. Half-steps that depend on contracts-that-don't-exist-yet don't get merged.

The original #04b was right for its moment — MIGRATION_DATABASE_URL had no consumer, so the refactor was a no-op. Today, with psycopg[binary]>=3.1 already in pyproject.toml, restoring the refactor would step into a broken state, not toward the goal. The split belongs in #03b.

Issue #21, opened 2026-07-17
Implement RiskAgent Node


pradyumna-001:
## Description
Implement the RiskAgent node inside `app/agents/risk.py` as outlined in the Phase 2 agentic workflow roadmap. This agent acts as a critical defensive layer in the multi-agent pipeline, reviewing up-stream data artifacts with explicit skepticism to identify hidden risk vectors, structural biases, and data gaps before final recommendations are formulated.

## Tasks
- Create the file `app/agents/risk.py`.
- Implement the entrypoint function `risk_agent_node(state: AgentState) -> AgentState` with clean logging context containing pipeline execution and morning note correlation IDs.
- Configure the agent execution block to use `gpt-4o` for complex adversarial reasoning and structural evaluation.
- Inject an explicit adversarial system prompt instructing the agent to operate as a deeply skeptical, hyper-critical financial analyst tasked with discovering overlooked risks, inconsistent assumptions, and optimism biases.
- Design the agent to ingest existing state context exclusively: it must read `macro_context`, `company_events`, and `quant_metrics` directly from the `AgentState`.
- Enforce strict boundaries: the RiskAgent MUST NOT fetch external data or trigger any search networks (e.g., Tavily, yfinance). It only critiques the context accumulated by upstream base agents.
- Parse the structured LLM output and populate `state['risk_flags']` with a list of `RiskFlag` Pydantic models, ensuring each flag captures a specific risk category, a probability scale, an impact scale, and a descriptive breakdown.
- Implement data gap amplification logic: if `state['flags']` contains active `DataFlag` warnings, the RiskAgent must explicitly evaluate those data omissions as separate, amplified threat factors in its risk profile.
- Create an isolated unit test suite verifying that the RiskAgent successfully identifies at least one valid, realistic risk factor when run against structured mock data.

### 1. Pure Analytical Isolation
The RiskAgent must operate entirely as an offline analyzer relative to external networks. No API clients other than the core OpenAI/OpenRouter chat completion gateway may be instantiated inside this node.

### 2. Additive State Invariant
The agent must append its newly identified risks to `state['risk_flags']` without wiping out or modifying data schemas generated by the `MacroAgent`, `CompanyAgent`, or `QuantAgent`.

### 3. Missing Structural Fields Handling
If an upstream field like `macro_context` or `quant_metrics` is missing or null (due to previous node failures or timeouts), the agent must gracefully handle the absence and log a specific data gap risk flag to alert downstream execution components.

---

[...]

Issue #19, opened 2026-07-17
Implement Integration Tests for Base Agents


pradyumna-001:
## Description
Create an integrated test suite inside tests/integration/test_agents.py to verify the sequential behavior, state mutations, and error handling thresholds across MacroAgent, CompanyAgent, and QuantAgent. These tests must run completely with mocked external network dependencies (Tavily, yfinance/B3, and LLM providers) to guarantee deterministic execution inside the CI pipeline.

## Tasks
- Create tests/integration/test_agents.py.
- Implement Integration Test 1: Mock Tavily API returning a 500 server error and verify that MacroAgent cleanly appends a high-priority DataFlag to the pipeline state without crashing.
- Implement Integration Test 2: Mock Tavily API returning a 500 server error and verify that CompanyAgent cleanly appends a high-priority DataFlag to the pipeline state without crashing.
- Implement Integration Test 3: Mock B3/yfinance data layers to return a 48-hour-old timestamp and verify that QuantAgent triggers its time guard, completely aborts calculations, and appends DataFlag('b3_api', 'data_outdated').
- Implement validation checks across all passing integration paths to ensure state['data_freshness'] is accurately updated with a real-time timestamp after each respective agent runs successfully.
- Enforce full monkeypatch / unittest.mock coverage across all OpenAI/OpenRouter chat completion calls to ensure zero live token consumption occurs during testing.
- Validate that the entire integration suite passes cleanly within the continuous integration (CI) run environment.

### 1. Zero External API Dependency
All integration tests must run fully offline. Any accidental real-world HTTP request made to OpenRouter, Tavily, or Yahoo Finance during the test run must fail the suite immediately to prevent fragile tests and unexpected API billings.

### 2. State Mutation Invariant
Verify that the shared AgentState dictionary accumulates data flags, data freshness markers, and data schemas correctly as it simulates flowing from one node to the next. Downstream structures must not erase or override unrelated fields set by upstream nodes.

### 3. Correlation Logging Verification
Ensure that simulated agent actions during integration steps continue to emit standard transaction logging contexts containing the mandatory pipeline_run_id and morning_note_id structural blocks.

---

## Expected Deliverables
- A standalone integration test file at tests/integration/test_agents.py containing 3 core green integration tests covering fail-visible network errors and stale data safeguards.
- 100% mocked external coverage passing perfectly under the local test runner and automated GitHub Actions CI infrastructure.

PR #40, opened 2026-08-04
feat(issue-04c): Closes #36 — POST /pipeline/trigger + SSE stub

### LINKED ISSUE #36: #04c — POST /pipeline/trigger and SSE stub for /morning-notes/{id}/stream
## Overview

Third chunk of issue #04. Adds the pipeline trigger endpoint and the
SSE stream stub. The SSE endpoint stays a stub in this issue; the
real event emission lands when issue #21 (SSE — atualizações em tempo
real) runs.

## Scope

1. **\`POST /pipeline/trigger\`** — returns 202 Accepted with a
   generated \`pipeline_run_id\` (UUID hex). Dispatches the ingestion
   pipeline as an async background task.
   - Note: actual agent implementation lands in issues #08–14. Here
     the endpoint only generates the run ID and enqueues; the
     background task is a stub that logs and exits.
2. **\`GET /morning-notes/{id}/stream\`** — stub SSE endpoint:
   - Returns \`text/event-stream\`.
   - Emits a single \`event: ready\ndata: {"morning_note_id": ...}\n\n\`
     event after connect, then closes.
   - Sets the right content-type and cache-control headers for SSE.
3. **Smoke probes** — extend \`tests/integration/\`:
   - \`POST /pipeline/trigger\` returns 202 with a UUID-shaped
     \`pipeline_run_id\`.
   - \`GET /morning-notes/{id}/stream\` returns the stub event and
     closes cleanly.

## Acceptance criteria

- [ ] \`POST /pipeline/trigger\` returns 202 with a non-empty UUID
  \`pipeline_run_id\`.
- [ ] \`GET /morning-notes/{id}/stream\` returns
  \`Content-Type: text/event-stream\` and emits the stub event.
- [ ] Both endpoints log via the #04b middleware (verify by checking
  the structured log output).
- [ ] \`mypy app/\` clean.
- [ ] \`pytest tests/integration/\` passes.

## Out of scope

- Real agent implementation (issues #08–14).
- Real SSE event emission (issue #21).
- \`POST /morning-notes/{id}/feedback\` (issue #28).
- App role split (#03b).
- RLS on remaining tables (#03a).

## How to verify

\`\`\`bash
cd finagent && uv run uvicorn app.main:app --port 8000
# In another shell:
curl -i -X POST http://localhost:8000/pipeline/trigger
# Expected: 202 Accepted with pipeline_run_id JSON.

curl -N http://localhost:8000/morning-notes/1/stream
# Expected: text/event-stream; one ready event; connection closes.

# Expected: previous tests + new trigger and stream probes pass.
\`\`\`

## Resolves

Third chunk of issue #04 per docs/issues.md.

pradyumna-001:
## What this PR does

- `app/schemas/pipeline.py`: `TriggerRequest`, `TriggerResponse` (server-minted UUIDs).
- `app/api/errors.py`: `PipelineError` base + `MorningNoteNotFound` + `InvalidTriggerPayload` typed subclasses.
- `app/services/pipeline.py`: `run_pipeline_stub` with contextvar-bound structured logging; module-level `_PIPELINE_REGISTRY` stub seam.
- `app/api/routes/pipeline.py`: `POST /pipeline/trigger` returns 202 with `TriggerResponse` body and `X-Pipeline-Run-Id` header; schedules stub via `BackgroundTasks`.
- `app/api/routes/morning_notes.py`: `GET /morning-notes/{id}/stream` returns `text/event-stream`, one `data:` event then close.
- `app/main.py`: wires the pipeline router; registers `MorningNoteNotFound` -> 404 exception handler.
- `tests/integration/test_pipeline.py`: 4 tests covering trigger 202, missing header 400, stream 200 + one event, stream 404.

## What this PR does NOT do

- Real LangGraph pipeline (lands in #14).
- Persistent `morning_notes` row write from the trigger (lands in #19; the registry is in-memory today).
- Multiple SSE event types (`agent_started`, etc.) -- single `data:` event only (lands in #21).
- ETag / Last-Event-ID / replay support.
- Authentication beyond the `manager-id` header.
- Logging of the 404 path inside the exception handler -- flagged for follow-up.

## How to verify

```bash
# Server (Postgres must be reachable; lifespan runs SELECT 1 at boot)
uvicorn app.main:app --reload

# Trigger -> expect 202 + body + X-Pipeline-Run-Id header
curl -i -X POST http://127.0.0.1:8000/pipeline/trigger \
  -H "manager-id: 1" -H "Content-Type: application/json" \
  -d '{"manager_id":1,"portfolio_id":42,"company_id":7}'

# Trigger without header -> expect 400
curl -i -X POST http://127.0.0.1:8000/pipeline/trigger \
  -H "Content-Type: application/json" \
  -d '{"manager_id":1,"portfolio_id":42,"company_id":7}'

# Stream a known note (paste morning_note_id from the first response)
curl -N -i "http://127.0.0.1:8000/morning-notes/<NOTE_ID>/stream"

# Stream unknown -> expect 404
curl -i "http://127.0.0.1:8000/morning-notes/00000000-0000-0000-0000-000000000000/stream"

# Tests
pytest tests/integration/test_pipeline.py -v
```

## Decisions worth flagging

- **`morning_note_id` is server-minted (Decision A1).** Client never supplies IDs; the server owns the namespace for its own resource lifecycle.
- **SSE payload is a single flat `data:` line (Decision B1).** No `event:` field -- YAGNI for #21 when the event taxonomy actually exists.
- **Local import of `_PIPELINE_REGISTRY` inside `trigger_pipeline`** to dodge a future circular import if `app/services/pipeline.py` ever adds a top-level reference to the routes module.
- **Tests bypass the DB fixtures** -- neither trigger nor stream touches Postgres today, so the `bound_engine` / `migrated_db_url` chain is unnecessary overhead.
- **Removed `app/api/routes.py`** (legacy `/health` + unused `/analyze/{ticker}`). `/health` is re-implemented in `app/main.py`; `/analyze/{ticker}` was dead code.
- **Commit history note:** one chunk commit has a typo (`errrors` instead of `errors`) left in place per the project's no-amend rule.


app/api/routes/pipeline.py
```diff
@@ -0,0 +1,38 @@
...
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="manager-id header required",
+        )
+
+    response = TriggerResponse()
+    from app.services.pipeline import _PIPELINE_REGISTRY
+    _PIPELINE_REGISTRY[str(response.morning_note_id)] = str(response.pipeline_run_id)
+    background.add_task(
+        run_pipeline_stub,
+        str(response.pipeline_run_id),
```
pradyumna-001, line 29: Q1 (architecture) — app/api/routes/pipeline.py:registry mutation: the route handler mutates a module-level `_PIPELINE_REGISTRY` dict that lives in another module (`app/services/pipeline.py`), and the stream route reads from the same dict via a top-level import. Two workers behind the same uvicorn process will share state fine, but two uvicorn workers won't — `_PIPELINE_REGISTRY[note_id]` written in worker A is invisible to worker B's stream handler. Why is the in-process dict acceptable for this PR, and at what point does that decision need to flip?

pradyumna-001: Single-worker uvicorn, in-process state survives across requests within one process. This is a stub that only needs to work under one worker, which is how we run it today. The in-proccess state is sized to the deployment that's actually running, not sized to a hypothetical future deployment. In the moment we run with more than one worker, we will need to stop using the dict and start using the Postgres.

app/api/routes/morning_notes.py
```diff
@@ -27,4 +35,19 @@ async def list_morning_notes(
             ), {"mid": manager_id},
         )
```
pradyumna-001, line 36: Q2 (architecture) — app/api/routes/morning_notes.py:media_type: the `StreamingResponse` ships with `media_type="text/event-stram"` (transposed letters — `r` and `s` swapped). Tests pass because the test only checks `status_code == 200` and reads body lines, never the `Content-Type` header. What's the impact of this typo today, and what's the cheapest way to make sure it doesn't ship to production unnoticed?

pradyumna-001: An EventSource will silently fail to connect if the Content-Type isn't text/event-stream. The test missed it because it only checks the status code. Since it's a one line fix, I will fix it right now.

app/main.py
```diff
@@ -1,11 +1,16 @@
...
+
 from fastapi import FastAPI
 from fastapi.responses import JSONResponse
+
 from sqlalchemy import text
 from sqlalchemy.exc import SQLAlchemyError
+
 from app.db.session import engine
 from app.middleware.correlation import CorrelationMiddleware
 from app.api.routes.morning_notes import router as morning_notes_router
```
pradyumna-001, line 11: Q3 (architecture) — app/main.py:exception_handler: `morning_note_not_found_handler` returns a 404 JSON response but doesn't log the exception. When #19 ships and real clients hit stale note_ids, an operator sees the access log line (`GET /morning-notes/X/stream → 404`) but not the exception itself. The PR body explicitly flags this as a follow-up. Why is "no log line" the right call for *this PR*, and what's the minimum log line that would make this handler debuggable without bloating the hot path?

pradyumna-001: Right now, the registry is empty unless this same process is triggered first. So during testing, "stream unknown" only happens via the test fixture with a hardcoded UUID. As for the minimum log line for the future, I think its safe to say we should log exc.args[0] (the note_id) and logger.warning (and if it's not enough we should log the full exception at logger.info). The 404 will become a hot path once #19 lands. We should evaluate the tradeoffs of info, warning and debug. That's a decision for the future, it's out of scope in the present PR.

app/services/pipeline.py
```diff
@@ -0,0 +1,24 @@
...
+
+from app.core.context import (
+    current_morning_note_id,
+    current_pipeline_run_id
+)
+
+logger = logging.getLogger(__name__)
+
+
+_PIPELINE_REGISTRY: dict[str, str] = {}
```
pradyumna-001, line 12: Q4 (architecture) — app/services/pipeline.py:contextvar reset: the stub wraps the work in `try/finally` and resets both `current_pipeline_run_id` and `current_morning_note_id` tokens. BackgroundTasks runs the stub in the same asyncio task that served the request, so the contextvars would be reset on task exit anyway — but if a future caller awaits this stub inside a larger pipeline (e.g., #14's LangGraph), the context would leak to whatever ran next. Is the `try/finally` defending against a current bug or a future bug, and what's the concrete scenario where the reset matters but task-exit wouldn't have caught it?

pradyumna-001: try/finally is defending against a future bug. Today, stub task exits promptly, but if inside a longer-lived pipeline task, and that pipeline task does more work after the sub returns, the contextvars set during the stub's execution wouls still be visible to later code in the same task. That's why it's important to explicitly reset at the boundary of "this stub's reponsaibility" and not at the boundary of "this task's lifetime". If it wasn't for that, try/finally would actually be redundant (as it is today), but since we now it's gonna be worthy in the near future and this is a scoped implementation for right now that doesn't require hard decisions, it's already implemented in this PR.

pradyumna-001:
Peer review — 4/4 solved.

- **Q1** (registry mutation, multi-worker): single-worker is the deployment reality; flip is at `--workers > 1` or #19's DB swap. ✅ Solved.
- **Q2** (`text/event-stram` typo): EventSource silently refuses to connect; one-line fix on the way. ✅ Solved.
- **Q3** (404 handler no-log): justified for the stub; `logger.warning(exc.args[0])` is the future minimum; level re-evaluation belongs in #19. ✅ Solved.
- **Q4** (contextvar reset): defending a future bug — task-vs-boundary distinction is the right architecture. ✅ Solved.

No changes requested. Decisions hold. Closing review.

Issue #44, opened 2026-08-06
#05d CI: full integration coverage (future)


pradyumna-001:
Sub-issue of #05 (Base unit tests + CI/CD). Future work — #05c ships unit-only CI; this issue adds the integration layer once #05c is stable.

## What lands

- Add `integration-tests` job to the same workflow (or create `.github/workflows/ci-integration.yml`)
  - Postgres service container (`postgres:16` with `pgvector` installed via init script)
  - Migration step before pytest
  - Run: `pytest tests/integration/ -v`
- Move `pg_container` fixture from testcontainers to Postgres-in-GHA (testcontainers-in-CI is flaky on shared runners)
- Add `services:` block in the job and `DATABASE_URL` pointing to `localhost:5432`
- Cache Docker layers to speed up subsequent runs
- Branch protection: mark `ci / integration-tests` as required check

## Verification

- Full pipeline runs in < 5 min
- Both unit and integration suites pass on a green push

## Deliverable

Integration tests run in CI on every push. Both unit and integration are required checks before merge.

Issue #43, opened 2026-08-06
#05c CI workflow — unit-only, green-on-push


pradyumna-001:
Sub-issue of #05 (Base unit tests + CI/CD). CI starts unit-only; integration coverage lands in #05d because Postgres-in-GHA setup is its own concern.

## What lands

- `.github/workflows/ci.yml`
  - Trigger: push on any branch + PR against `main`
  - Setup: `actions/checkout@v4`, `actions/setup-python@v5` with Python 3.11, `pip install -e ".[dev]"`
  - Run: `ruff check .", `mypy app`, `pytest tests/unit/ -v`
  - Cache pip based on `uv.lock` (decide in chunk)
- Branch protection rule: "Require status checks to pass before merging" — mark `ci / unit-tests` as required

## Verification

- Push a commit with a red test → CI fails, merge blocked
- Push a commit with a green test → CI passes

## Deliverable

Green CI on every push. Red CI blocks merge.

Issue #42, opened 2026-08-06
#05b DataFlag frozen dataclass + helper methods


pradyumna-001:
Sub-issue of #05 (Base unit tests + CI/CD). DataFlag is the type that #07 (Typed AgentState) will annotate on `state["flags"]`. Defining it now lets #07 import a stable shape and gives helpers like `.is_fatal()` from day 1.

## What lands

- `app/utils/flags.py` — `@dataclass(frozen=True) class DataFlag`
  - Fields: `source: str`, `severity: Severity` (enum: INFO / WARNING / FATAL), `message: str`, `created_at: datetime` (default factory)
- Methods: `.is_fatal() -> bool`, `.is_warning() -> bool`, `.to_dict() -> dict` (for JSONB serialization)
- `__post_init__` validates `message` non-empty and `source` non-empty
- `tests/unit/test_dataflag.py` with `test_dataflag_generation`
  - Source failure generates DataFlag with `severity=WARNING`, `source="tavily"`, message containing the failure
  - `__post_init__` rejects `source=""` with `ValueError`
  - `.to_dict()` serializes datetime as ISO-8601

## Constraints

- No dependency on #07 — #05b defines the shape that #07 consumes
- Pure Python, no Docker fixtures

## Deliverable

DataFlag dataclass importable from `app.utils.flags`. Test `test_dataflag_generation` passes.

Issue #10, opened 2026-07-14
arch: design and implement unified prompt management service and schemas


pradyumna-001:
## Description
Refactor how prompts are managed, validated, and loaded across the application. Instead of agent nodes reading text files directly from the disk using raw file I/O, we should leverage the empty `app/prompts` structure (`schemas/`, `services/`, and `main.py`) to build a unified Prompt Management Subsystem.

## Tasks
- [ ] **Define Input Schemas (`app/prompts/schemas/`):** Define Pydantic models to validate the inputs required by each agent prompt template (e.g., ensuring `search_results` is present and non-empty).
- [ ] **Build Prompt Service (`app/prompts/services/`):** Create a robust `PromptLoaderService` responsible for reading prompt files, caching them in memory to prevent repetitive disk reads, and formatting templates securely using Pydantic schemas.
- [ ] **Expose Service interface (`app/prompts/main.py`):** Provide clean, high-level functions or a singleton client class to retrieve and format prompts easily from within agent nodes.
- [ ] **Refactor Existing Agents:** Update the `MacroAgent` (and any other agents implemented by then) to use the new service instead of raw `open()` calls.

## Definition of Done
- All prompts are loaded and validated through the `app/prompts` service.
- Disk reads are cached or optimized.
- Existing tests pass cleanly.

PR #45, opened 2026-08-06
feat(issue-05a): Closes #41 — Unit-test base: freshness + confidence threshold

### LINKED ISSUE #41: #05a Unit-test base: freshness + confidence threshold
Sub-issue of #05 (Base unit tests + CI/CD). Decision on 2026-08-05: separate modules per responsibility instead of a single utils.py.

## What lands

- `app/utils/__init__.py` (namespace marker)
- `app/utils/freshness.py` — `is_data_fresh(source: str, fetched_at: datetime, *, now: datetime | None = None, max_age: timedelta = timedelta(hours=24)) -> bool`
- `app/utils/confidence.py` — `confidence_flag(score: float, *, threshold: float = 0.75) -> bool` (returns bool; #05b wraps in DataFlag)
- `tests/unit/test_freshness.py` with `test_data_freshness_check` (data > 24h returns False; fresh returns True; boundary at exactly 24h returns True)
- `tests/unit/test_confidence.py` with `test_confidence_threshold` (score < 0.75 returns flag; score == 0.75 returns flag; score > 0.75 returns False)

## Constraints

- Pure Python — no Docker fixtures, no `conftest.py` import
- No dependency on `DataFlag` (#05b) — `confidence_flag` returns `bool`, not DataFlag

## Deliverable

Two modules + two test files. `pytest tests/unit/ -v` passes locally.

pradyumna-001:
## What this PR does

- Adds `app/utils/__init__.py` (empty package marker).
- Adds `app/utils/freshness.py` exporting `is_data_fresh(source, fetched_at, *, now=None, max_age=timedelta(hours=24)) -> bool`.
- Adds `app/utils/confidence.py` exporting `confidence_flag(score, *, threshold=0.75) -> bool`.
- Adds `tests/unit/test_freshness.py` with `test_data_freshness_check` (3 assertions: 1h fresh, 24h boundary fresh, 48h stale).
- Adds `tests/unit/test_confidence.py` with `test_confidence_threshold` (3 assertions: 0.5 flagged, 0.75 boundary flagged, 0.9 not flagged).
- All tests pass locally: `uv run pytest tests/unit/ -v` → `4 passed in 0.65s` (2 new + 2 pre-existing correlation tests).
- Lint clean: `uv run ruff check app/utils tests/unit` → All checks passed.
- Types clean: `uv run mypy app/utils/` → Success: no issues found in 3 source files.

## What this PR does NOT do

- Does not define `DataFlag` — that lands in #05b (#42).
- Does not return `DataFlag` from `confidence_flag`. The function returns `bool` to keep #05a independent of #05b's not-yet-defined type. The refactor to `-> DataFlag` happens in #05b with a ~6-line diff (import, signature, return statement, test assertions).
- Does not introduce a `conftest.py`, fixtures, or any testcontainers. Pure-Python tests only, per the issue spec.
- Does not add a CI workflow file — that's #05c (#43).
- Does not address integration coverage — that's #05d (#44).

## How to verify

```bash
# 1. Sync the branch
git fetch origin
git checkout feature/issue-05-unit-tests-ci

# 2. Run the unit suite — expect 4 passed
uv run pytest tests/unit/ -v

# 3. Lint check — expect "All checks passed"
uv run ruff check app/utils tests/unit

# 4. Type check — expect "Success: no issues found"
uv run mypy app/utils/

# 5. Targeted boundary check (proves the <=\u00a0invariant on is_data_fresh)
uv run python -c "from datetime import datetime, timedelta; from app.utils.freshness import is_data_fresh; n = datetime(2026, 8, 5, 12); print(is_data_fresh('b3', n - timedelta(hours=24), now=n))"
# Expected output: True
```
## Decisions worth flagging
1. Boundary semantics: <= at exactly max_age / threshold. Stale-or-borderline data triggers the flag at max_age + epsilon, not at the boundary itself. This is the project's "Fail Visible" invariant: a flag at the edge is fine, but a silent pass at the edge is not. The issue spec is explicit ("boundary em exatamente 24h retorna True"); we honor it.
2. Keyword-only now and threshold via *,. Both functions have multiple optional args of similar type (datetime | None / timedelta; float / float). The * forces keyword-only access so a caller can't silently bind a value to the wrong parameter. Verified by a bonus call in the verify step (confidence_flag(0.8, threshold=0.85)) — without *, that positional override would succeed with no error.
3. Unused source parameter in is_data_fresh is intentional. The body never reads it today; ruff's default rule set does not flag unused function arguments (the ARG family is opt-in). The parameter is reserved for #05b's DataFlag.source wiring — when is_data_fresh is later rewritten to return a DataFlag instead of a bool, source will populate the flag's source field. Keeping it in the signature now avoids breaking callers (the QuantAgent in #10) when the refactor lands.
4. confidence_flag returns bool, not DataFlag. Decision explicitly flagged in the issue spec ("ou retorna DataFlag direto — decidir no chunk"). We chose bool to honor the issue's "Sem dependência do #07" rule symmetrically — #05a must not depend on #05b's not-yet-defined type. The refactor is small (~6 lines) and lands in #05b.
5. from __future__ import annotations deliberately omitted. pyproject.toml declares requires-python = ">=3.11", so PEP 604 syntax (X | None) is native. The future-import is redundant noise at this version floor.

### DIFF
--- a/app/utils/confidence.py
+++ b/app/utils/confidence.py
@@ -1,0 +1,2 @@
+def confidence_flag(score: float, *, threshold: float = 0.75) -> bool:
+    return score <= threshold
--- a/app/utils/freshness.py
+++ b/app/utils/freshness.py
@@ -1,0 +1,7 @@
+from datetime import datetime, timedelta
+
+
+def is_data_fresh(source: str, fetched_at: datetime, *, now: datetime | None = None, max_age: timedelta = timedelta(hours=24)) -> bool:
+    if now is None:
+        now = datetime.now()
+    return (now - fetched_at) <= max_age
--- a/tests/unit/test_confidence.py
+++ b/tests/unit/test_confidence.py
@@ -1,0 +1,7 @@
+from app.utils.confidence import confidence_flag
+
+
+def test_confidence_threshold() -> None:
+    assert confidence_flag(0.5) is True
+    assert confidence_flag(0.75) is True
+    assert confidence_flag(0.9) is False
--- a/tests/unit/test_freshness.py
+++ b/tests/unit/test_freshness.py
@@ -1,0 +1,15 @@
+from datetime import datetime, timedelta
+
+from app.utils.freshness import is_data_fresh
+
+FIXED_NOW = datetime(2026, 8, 5, 12, 0, 0)
+
+
+def test_data_freshness_check() -> None:
+    fresh = is_data_fresh("b3", FIXED_NOW - timedelta(hours=1), now=FIXED_NOW)
+    boundary = is_data_fresh("b3", FIXED_NOW - timedelta(hours=24), now=FIXED_NOW)
+    stale = is_data_fresh("b3", FIXED_NOW - timedelta(hours=48), now=FIXED_NOW)
+
+    assert fresh is True
+    assert boundary is True
+    assert stale is False


pradyumna-001:
The `source` parameter on `is_data_fresh` is unused in this PR. The PR description defends it as reserved for #05b's `DataFlag.source` wiring. Confirming for the record: when #05b lands and refactors `is_data_fresh` to return a `DataFlag`, this `source` arg will populate `DataFlag.source` — yes/no?

PR #46, opened 2026-08-06
feat(issue-05b): Closes #42 — DataFlag frozen dataclass + Severity enum

### LINKED ISSUE #42: #05b DataFlag frozen dataclass + helper methods

pradyumna-001:
## What this PR does

- Adds `app/utils/flags.py` exporting `Severity` enum (INFO / WARNING / FATAL) and `DataFlag` frozen dataclass with fields `source`, `severity`, `message`, `created_at` (default factory).
- `DataFlag.__post_init__` rejects empty `source` and empty `message` with `ValueError`.
- Methods: `is_fatal()`, `is_warning()`, `to_dict()`. `to_dict()` serializes `severity` via `.value` and `created_at` via `.isoformat()` for Postgres JSONB.
- Adds `tests/unit/test_dataflag.py` with 4 tests: WARNING flag generation, empty-source rejection, empty-message rejection, `.to_dict()` ISO-8601 round-trip with `before/after` bounds.
- All tests pass locally: `uv run pytest tests/unit/ -v` → 4 passed.
- Lint clean: `uv run ruff check app/utils tests/unit` → All checks passed.
- Format clean: `uv run ruff format --check app/utils tests/unit` → already formatted.

## What this PR does NOT do

- Does not return `DataFlag` from `confidence_flag` (lands as a follow-up to #05a now that the type exists — small refactor, ~6 lines).
- Does not annotate `state["flags"]` in `AgentState` — that's #07.
- Does not add a CI workflow file — that's #05c (#43).
- Does not address integration coverage — that's #05d (#44).
- Does not add `is_info()` — no caller branches on INFO today (YAGNI; add when needed).

## How to verify

```bash
# 1. Sync the branch
git fetch origin
git checkout feature/issue-05b-dataflag

# 2. Run the new tests — expect 4 passed
uv run pytest tests/unit/test_dataflag.py -v

# 3. Run the full unit suite — expect prior 4 + new 4 = 8 passed
uv run pytest tests/unit/ -v

# 4. Lint check — expect "All checks passed"
uv run ruff check app/utils tests/unit

# 5. Format check — expect "already formatted"
uv run ruff format --check app/utils tests/unit

# 6. Smoke: construct a WARNING flag and inspect its dict form
uv run python -c "from app.utils.flags import DataFlag, Severity; f = DataFlag(source='tavily', severity=Severity.WARNING, message='fetch failed: 500'); print(f.is_warning(), f.is_fatal(), f.to_dict())"
# Expected: True False {'source': 'tavily', 'severity': 'warning', 'message': 'fetch failed: 500', 'created_at': '...'}
```

## Decisions worth flagging

1. **Severity uses `.value` ("info"/"warning"/"fatal"), not `.name`.** `.value` is the portable contract: renaming the enum member later (e.g., FATAL → CRITICAL) doesn't break stored JSONB values. `.name` is a Python identifier and is allowed to drift; `.value` is not.
2. **frozen=True gives hashability for free.** `state["flags"]: list[DataFlag]` may need dedup via `set()` later. `frozen=True` generates `__hash__` from the fields automatically; a mutable dataclass has `__hash__ = None` and can't be set-deduped. The cost is zero unless someone tries to assign post-init, which `@dataclass` then blocks — desirable here.
3. **No timezone-aware default.** `field(default_factory=datetime.now)` produces naive datetimes. Mixing naive and aware `datetime` objects raises `TypeError` on comparison, so the entire module is naive-by-default. Adding `datetime.UTC` later is a migration across all flag creation sites, not a one-line change. Decision flagged so future-you doesn't quietly mix.
4. **`is_info()` deliberately omitted.** No agent in #08/#09/#10 branches on INFO; it would be load-bearing only if a future caller needed it. Public surface stays minimal until demand arrives.
5. **`__post_init__` uses two separate `if not x: raise` blocks**, not a single combined check. Each error message is specific to its field. A generic "validation failed" message would force callers (and the tests) to grep the exception text — current form makes the contract self-documenting.
6. **Trailing comma in `to_dict` literal is intentional** — ruff format requires it for multi-line dicts. Caught and fixed by `ruff format` before commit.

### DIFF
--- a/app/utils/flags.py
+++ b/app/utils/flags.py
@@ -1,0 +1,37 @@
+from dataclasses import dataclass, field
+from datetime import datetime
+from enum import Enum
+
+
+class Severity(Enum):
+    INFO = "info"
+    WARNING = "warning"
+    FATAL = "fatal"
+
+
+@dataclass(frozen=True)
+class DataFlag:
+    source: str
+    severity: Severity
+    message: str
+    created_at: datetime = field(default_factory=datetime.now)
+
+    def __post_init__(self) -> None:
+        if not self.source:
+            raise ValueError("DataFlag.source must be non-empty")
+        if not self.message:
+            raise ValueError("DataFlag.message must be non-empty")
+
+    def is_fatal(self) -> bool:
+        return self.severity == Severity.FATAL
+
+    def is_warning(self) -> bool:
+        return self.severity == Severity.WARNING
+
+    def to_dict(self) -> dict[str, object]:
+        return {
+            "source": self.source,
+            "severity": self.severity.value,
+            "message": self.message,
+            "created_at": self.created_at.isoformat(),
+        }
--- a/tests/unit/test_dataflag.py
+++ b/tests/unit/test_dataflag.py
@@ -1,0 +1,40 @@
+from datetime import datetime
+
+import pytest
+
+from app.utils.flags import DataFlag, Severity
+
+
+def test_dataflag_generation() -> None:
+    flag = DataFlag(
+        source="tavily", severity=Severity.WARNING, message="fetch failed: 500"
+    )
+    assert flag.source == "tavily"
+    assert flag.severity is Severity.WARNING
+    assert "fetch failed" in flag.message
+    assert flag.is_warning() is True
+    assert flag.is_fatal() is False
+
+
+def test_dataflag_rejects_empty_source() -> None:
+    with pytest.raises(ValueError, match="source must be non-empty"):
+        DataFlag(source="", severity=Severity.INFO, message="x")
+
+
+def test_dataflag_rejects_empty_message() -> None:
+    with pytest.raises(ValueError, match="message must be non-empty"):
+        DataFlag(source="tavily", severity=Severity.WARNING, message="")
+
+
+def test_to_dict_serializes_datetime_as_iso_8601() -> None:
+    before = datetime.now()
+    flag = DataFlag(source="b3", severity=Severity.FATAL, message="stale")
+    payload = flag.to_dict()
+    after = datetime.now()
+
+    assert payload["source"] == "b3"
+    assert payload["severity"] == "fatal"
+    assert payload["message"] == "stale"
+
+    parsed = datetime.fromisoformat(payload["created_at"])
+    assert before <= parsed <= after


app/utils/flags.py
```diff
@@ -0,0 +1,37 @@
+from dataclasses import dataclass, field
+from datetime import datetime
+from enum import Enum
+
+
+class Severity(Enum):
+    INFO = "info"
+    WARNING = "warning"
+    FATAL = "fatal"
```
pradyumna-001, line 9: Q2 (architecture) — app/utils/flags.py:9 frozen=True. If a future caller (e.g., an agent retrying with extra context) wants to update the message without rebuilding the flag, what's the path? Does that path exist today, or is rebuilding the dataclass the only option — and is that intentional?

app/utils/flags.py
```diff
@@ -0,0 +1,37 @@
+from dataclasses import dataclass, field
+from datetime import datetime
+from enum import Enum
+
+
+class Severity(Enum):
```
pradyumna-001, line 6: Q1 (architecture) — app/utils/flags.py:6 Severity.INFO='info' lowercase. Why not uppercase like the Python enum name (INFO / WARNING / FATAL)? What's the contract — does the wire/JSONB value have to stay lowercase, and who depends on that?

app/utils/flags.py
```diff
@@ -0,0 +1,37 @@
...
+    INFO = "info"
+    WARNING = "warning"
+    FATAL = "fatal"
+
+
+@dataclass(frozen=True)
```
pradyumna-001, line 12: Q3 (architecture) — app/utils/flags.py:12 created_at default_factory=datetime.now. If a flag is constructed in tests with frozen time and later serialized via to_dict(), does the ISO-8601 round-trip preserve the timezone contract, or does it drift to naive local time on every call? Walk me through the case where someone passes their own datetime explicitly.

tests/unit/test_dataflag.py
```diff
@@ -0,0 +1,40 @@
+from datetime import datetime
+
+import pytest
+
```
pradyumna-001, line 4: Q4 (architecture) — tests/unit/test_dataflag.py:4 plain 'from app.utils.flags import DataFlag, Severity'. No conftest, no fixture, no shared helper. If #05a+#05b grow into 5+ files in app/utils/, does each test file keep its own imports, or is there a planned shared module? What's the convention you're locking in here?

app/utils/flags.py
pradyumna-001: Lowercase vales are the wire  format, they live in JSONB after to_dict(). Python enum name is the source identifier, the value is the serialized identifier. They drift on purpose so I can rename the enum later without migrating JSONB rows.

app/utils/flags.py
pradyumna-001: Rebuild-only, intentional. A DataFlag is a arecord of an observation at created_at: once emitted, it shouldn't change. Mutable would break hashability, and I'd rather lose the ergonomic edit than lose set-based dedup. Altough later I could add a with_message() helper if necessary.

app/utils/flags.py
pradyumna-001: Naive-by-default. The whole flags.py module produces naive datetimes — default_factory=datetime.now is naive, so any caller that doesn't pass created_at explicitly gets naive. The trade-off: in CI (UTC) vs. prod (BRT-3), naive datetime.now() returns different wall-clock strings for the same instant. We accept that for now because there's only one Python process creating flags, and the JSONB column treats it as opaque string anyway. If we ever need to compare timestamps across processes or across timezones, the migration is 'add tz=datetime.UTC to every datetime.now() call in this module' — one grep, one sed. Until that need shows up, naive is the simpler default.

tests/unit/test_dataflag.py
pradyumna-001: Per-file imports, no conftest, no helpers. Each test file is self-contained — if you read test_dataflag.py, you see exactly what shape a DataFlag has in the test. Pulling constructors into conftest.py would save maybe 2 lines per test while hiding the data shape that the test is actually checking. The cost of locality (slight duplication) is lower than the cost of hidden state (read the helper to understand the test). I'll add a conftest the day two test files need the same fixture with the same parameters — not before.

PR #64, opened 2026-08-13
refactor(tavily): Closes #62 — extract TavilyService class

### LINKED ISSUE #61: Audit: MacroAgent test + agent bugs found — source string + latent parse bug + prompt regressions
# Follow-up: MacroAgent audit — agent code + test suite

Filed during #09 CompanyAgent work (session 2026-08-12) when Prady asked: "is macro actually perfect? maybe we should make it better."

Audit covers `app/agents/macro.py` (97 lines) and `tests/unit/test_macro_agent.py` (121 lines), cross-referenced against the contracts we just locked in #09 CompanyAgent (`app/agents/company.py`, `app/prompts/company.py`, `tests/unit/test_company_agent*.py`).

## Summary

**Macro has 6 distinct issues** — 2 are real bugs (one in agent code, one in tests), 4 are thin/shallow patterns that should match the company standard. None are pipeline-breaking; all are quality regressions vs the #09 work.

### 1. **BUG (agent code + test) — `source="tavily"` for LLM failure (HIGH)**

**Location:** `app/agents/macro.py:78–85` (agent code) and `tests/unit/test_macro_agent.py:107` (test assertion).

In `macro.py`, the LLM-None branch (`if summary is None:`) constructs a `DataFlag` with `source="tavily"`. That's wrong: the LLM (NVIDIA NIM) failed, not Tavily. Tavily succeeded — it returned the data we asked for. The flag source should be `"nvidia_nim"`, exactly as we locked in `company.py`.

The test at `test_macro_agent.py:107` asserts `flag.source == "tavily"` — it was written against the buggy agent code, so the assertion *passes* but the contract is wrong. Fixing the agent code without fixing the test would make the test fail; both must be fixed together.

**Why it matters:** downstream `RiskAgent` (#12) and operator triage both depend on the `source` field to know *which upstream* failed. A mislabeled source sends the operator to file a ticket against Tavily when they should be filing one against NVIDIA NIM (or our LLM call site).

**Fix:** `app/agents/macro.py:81` change `source="tavily"` → `source="nvidia_nim"`. `tests/unit/test_macro_agent.py:107` change `assert flag.source == "tavily"` → `assert flag.source == "nvidia_nim"`. Done.

---

### 2. **BUG (test) — `test_macro_agent_tavily_failure` asserts shallow FATAL/WARNING contract (MEDIUM)**

**Location:** `tests/unit/test_macro_agent.py:68–72`.

The HTTP-failure test only asserts:
- `result["macro_context"] is None` ✓
- `flag.source == "tavily"` ✓
- `isinstance(result["data_freshness"]["macro"], datetime)` ✓

It never asserts `flag.severity == Severity.WARNING`. Compare to the company tests, which explicitly assert `Severity.FATAL` on the no-key test and on the all-null-fields test. If a regression changes the HTTP-failure branch from `WARNING` to `FATAL` (or to `INFO`), this test passes silently.

**Fix:** import `Severity` (already imported via `DataFlag` chain — confirm) and add `assert flag.severity == Severity.WARNING` after the existing `flag.source` assertion.

---

### 3. **LATENT BUG (agent code) — parse block has no try/except validator (HIGH)**

**Location:** `app/agents/macro.py:65–93`.

Same latent bug we discovered and fixed in #09 `company.py`: the `if data and data.get("results"):` block does unconditional `top.get(...)` and ships garbage to `summarize()` if Tavily returns `{"results": [{"title": null, "content": null}]}`. No `ValueError` is raised; the LLM gets `"\n\n\n"` or `"None\n\nNone\nNone"`, returns garbage, and the morning note ships garbage with no flag.

We just redesigned this for company: a validator gate (`if not (top.get("title") and top.get("content")):`) wrapped in `try/except (ValueError, KeyError, IndexError)` → `FATAL` `DataFlag(source="tavily", message=f"parse-block failed for {ticker}: {exc}")`. The macro version needs the identical refactor.

Note: macro doesn't have a `company_ticker` state field. The message format should be `f"parse-block failed: {exc}"` (no ticker) — since there's only one "Brazil macro" query per pipeline run, the ticker isn't needed for routing.

**Fix:** backport the company.py fix-chunk we just wrote (lines 76–) to macro.py, dropping the ticker interpolation. Add an all-null-fields test (`test_macro_agent_all_null_fields`) mirroring `test_company_agent_all_null_fields`.

---

### 4. **STYLE (prompts module) — `MACRO_PROMPTS.build_user_prompt` lacks the company improvements (MEDIUM)**

**Location:** `app/prompts/macro.py:8–12`.

`company.py`'s `build_user_prompt` was written yesterday with three improvements over macro:
- **(a) f-string instead of `+` concatenation:** `f"Summarize...{raw_text}"` instead of `"... :\n\n" + raw_text`.
- **(b) Docstring with `Args`/`Returns`/`Raises`.**
- **(c) `if not raw_text: raise ValueError(...)` guard clause.**

Macro has none of these. The `+` concatenation is benign; the missing docstring is a docs gap; the missing `ValueError` guard means the macro parse-block fix-chunk (#3 above) can't exist — we need the prompt to raise on empty raw_text so the parse-block `try/except` catches something.

**Fix:** backport all three improvements from `app/prompts/company.py` to `app/prompts/macro.py`. Specifically:
- Add `if not raw_text: raise ValueError("raw_text must be non-empty")` at the top of `build_user_prompt`.
- Convert `+` to f-string.
- Add docstring matching `company.py`'s shape.

Update `"3-5 sentences"` → macro can keep `"concise paragraph"` (different output shape — macro is a paragraph, company is event bullets) OR standardize on the more specific form. That's an editorial call, not a code-correctness one.

---

### 5. **STYLE (test) — `make_state()` helper vs inline `create_initial_state(...)` (LOW)**

**Location:** `tests/unit/test_macro_agent.py:10–16`.

Macro uses a local `make_state()` helper at module scope; company inlines `create_initial_state(manager_id=1, company_ticker="PETR4", pipeline_run_id=..., morning_note_id=...)` per test. Both work. Macro is more DRY (single edit point); company is more grep-able (search finds each test).

This is a judgment call, not a bug. Pick one and standardize across both macro and company test files.

**Fix:** decision call. Recommendation: keep `make_state()` helper — it's the smaller diff (macro stays, company refactors). Or keep inline (macro refactors to inline) if grep-ability wins. Either way, *both files should match*.

---

### 6. **STYLE (test) — `flag.source == "tavily"` assertion in `test_macro_agent_tavily_failure` (LOW)**

**Location:** `tests/unit/test_macro_agent.py:71`.

The HTTP-failure test correctly asserts `flag.source == "tavily"` — that's fine (Tavily did fail). But it doesn't assert `flag.severity` (issue #2 above). This is mostly a sub-point of #2 — calling out separately so the fix PR covers both in one pass.

Also: the HTTP-failure branch sets `severity=Severity.WARNING`. The test should explicitly assert that — not just `isinstance(flag, DataFlag)`. Same fix as #2.

---

## What is NOT a bug

- `datetime.now(UTC)` instead of deprecated `utcnow()` — already fixed in #08 (yesterday's issue). Good.
- `httpx.AsyncClient(timeout=10.0)` pattern — matches company. Good.
- `include_domains` filter on Tavily query — already added in #08. Good.
- `MacroOutput` TypedDict — already in #08. Good.
- Correlation-ID entry logger (`extra={pipeline_run_id, morning_note_id, manager_id}`) — already in #08. Good.
- The `summarize` `None`-fallback to raw content — semantic is right, only the `source` field is mislabeled (issue #1 above).

## Why not fixed in #09 PR

Scoping discipline: #09 is CompanyAgent. #08 (macro) closed yesterday with a green suite — but the audit wasn't part of #08's exit criteria. Filing this as a separate issue keeps the PR scope clean.

## Scope of the follow-up PR

Minimum viable fix (bug + critical latent):
1. `app/agents/macro.py:81` — `source="tavily"` → `source="nvidia_nim"` (issue #1).
2. `tests/unit/test_macro_agent.py:107` — `assert flag.source == "tavily"` → `"nvidia_nim"` (issue #1).
3. Backport company.py parse-block fix-chunk to `app/agents/macro.py:65–93` — validator + try/except + FATAL flag (issue #3).
4. Add `test_macro_agent_all_null_fields` test mirroring `test_company_agent_all_null_fields` (issue #3).
5. Backport company.py `build_user_prompt` improvements to `app/prompts/macro.py` — ValueError guard, f-string, docstring (issue #4).

Polish pass (low-priority):
6. `tests/unit/test_macro_agent.py:71` — add `assert flag.severity == Severity.WARNING` (issues #2 + #6).
7. Standardize `make_state()` helper vs inline `create_initial_state(...)` across macro + company test files (issue #5).

## Suggested branch
`fix/issue-08b-macro-audit-fixes`

## Suggested priority
- **Issues #1, #3, #4** (bug + latent + dependency): HIGH — file before #12 RiskAgent work starts (RiskAgent consumes `flag.source`, will inherit the mislabel bug).
- **Issue #2** (shallow FATAL/WARNING assertion): MEDIUM — quality, not pipeline-affecting.
- **Issues #5, #6** (style standardization): LOW — cleanup, no user impact.

## Found by
Session 2026-08-12, issue #09 work. Prady asked "is macro actually perfect?" after we audited company tests against macro's precedent. The audit surfaced 6 issues; Socratic protocol on "macro's source='tavily' for LLM" revealed the test bug; deeper read of `macro.py` parse block revealed the latent bug (same root cause as company's, not yet fixed); comparison to `app/prompts/company.py` revealed the prompt-module regressions.

### LINKED ISSUE #62: Refactor: Extract TavilyService — deduplicate HTTP + parse + source-string discipline across agents
# Follow-up: Extract TavilyService — deduplicate HTTP + parse + source-string discipline

Filed during #09 CompanyAgent work (session 2026-08-12) just before the PR went up for review. Prady asked: "does it make sense to create a future issue to develop a tavily service module?" — the answer was yes, after surfacing two real architectural tensions.

## Why now

Two call sites today (`app/agents/macro.py`, `app/agents/company.py`) already duplicate:
- `httpx.AsyncClient(timeout=10.0)` context manager.
- `client.post("https://api.tavily.com/search", headers={"Authorization": f"Bearer {tavily_key}"}, json=payload)`.
- `except Exception as exc:` → `WARNING` `DataFlag(source="tavily", message=f"Tavily request error: {exc}")`.
- `if not tavily_key:` → `FATAL` `DataFlag(source="tavily", message="TAVILY_API_KEY missing")` guard.
- The parse-block guard we just added in #09 (`try/except (ValueError, KeyError, IndexError)` → `FATAL`), soon to be backported to `macro.py` per issue #61.

Future agents `risk_agent` (#12), `quant_agent` (#13) will need news/event data — more duplication.

### 1. Source-string discipline
Today, `source="tavily"` vs `source="nvidia_nim"` lives in each agent's local code. We already found `macro.py` uses `source="tavily"` for LLM-None failures — a mislabel (filed in #61). A `TavilyService.search()` that returns `(results, error_flag)` centralizes "Tavily succeeded/failed"; the caller attaches LLM-specific `source="nvidia_nim"` only when the LLM fails. The bug class disappears.

### 2. Parse-block guard
The `try/except (ValueError, KeyError, IndexError)` → `FATAL` guard we just added to `company.py` (lines 76–) needs backporting to `macro.py` per #61. Next week `risk.py` will need the same. A service that returns *typed* `TavilyResult` (`results: list[TavilyArticle]`, `error: DataFlag | None`) instead of raw dicts pushes the guard *inside* the service. Agents stop touching `dict.get("title")` and receive `article.title: str | None`.

## Proposed shape

```python
# app/services/tavily.py
from dataclasses import dataclass
from app.utils.flags import DataFlag, Severity

@dataclass(frozen=True)
class TavilyArticle:
    title: str | None
    content: str | None
    url: str | None

@dataclass(frozen=True)
class TavilyResult:
    articles: list[TavilyArticle]
    error: DataFlag | None  # non-None if HTTP/parse/key failed

class TavilyService:
    def __init__(self, api_key: str, timeout: float = 10.0) -> None: ...
    async def search(
        self,
        query: str,
        include_domains: list[str],
        max_results: int = 10,
    ) -> TavilyResult: ...
```

### Rewrite of `company_agent_node` call site (sketch)

```python
tavily = TavilyService(api_key=settings.TAVILY_API_KEY)
if not tavily.api_key:
    state["flags"].append(DataFlag(source="tavily", severity=FATAL, message="TAVILY_API_KEY missing"))
    state["company_events"] = []
    state["data_freshness"]["company"] = datetime.now(UTC)
    return state

result = await tavily.search(
    query=f"{ticker} news Brazil",
    include_domains=["cvm.gov.br", "infomoney.com.br", "globo.com/valor-economico"],
    max_results=10,
)
if result.error:
    state["flags"].append(result.error)
    state["company_events"] = []
    state["data_freshness"]["company"] = datetime.now(UTC)
    return state

top = result.articles[0]
raw_text = f"{top.title or ''}\n\n{top.content or ''}\n{top.url or ''}"
# parse-block try/except no longer needed — service returned typed articles
summary = await summarize(system=COMPANY_PROMPTS.system, user=COMPANY_PROMPTS.build_user_prompt(raw_text=raw_text))
...
```

Notes:
- The `if not tavily.api_key` guard could move *inside* `TavilyService.__init__` and `search()` returns `TavilyResult(articles=[], error=FATAL_flag)` — but that hides the missing-key check behind a method call. Keeping the explicit guard in the agent keeps the contract visible at the boundary. The shape decision is open for debate in the issue.
- `TavilyArticle.title` is `str | None`. The agent still does `.title or ''` for the f-string — but `top.title` is now a typed field, not a dict access; the `.get("title")` and dict-shape-question-marks disappear.
- The `summarize-is-None` branch stays in the agent (LLM failure is agent responsibility, not service responsibility). The `source="nvidia_nim"` flag stays correct.

## Scope of the refactor PR

1. New file `app/services/tavily.py` with `TavilyArticle`, `TavilyResult`, `TavilyService`.
2. Unit tests for `TavilyService.search()`:
   - happy path returns typed articles;
   - HTTP failure returns `error: DataFlag(source="tavily", severity=WARNING)`;
   - all-null-fields response returns `error: DataFlag(source="tavily", severity=FATAL)` with disambiguating message;
   - empty key handled (either in `__init__` or in `search` — design call).
3. Rewrite `app/agents/macro.py` to use `TavilyService`. Drop local httpx boilerplate. Delete the parse-block try/except (service owns it). Keep LLM-None branch + `source="nvidia_nim"` flag.
4. Rewrite `app/agents/company.py` to use `TavilyService`. Same deletions as macro.
5. Update existing unit tests (`test_macro_agent.py`, `test_company_agent*.py`) to mock `TavilyService.search()` instead of `httpx.AsyncClient`. Tests get simpler — no `mock_client.__aenter__.return_value = mock_client` plumbing.
6. Update `scripts/run_macro_agent.py` and `scripts/run_company_agent.py` (after sys.path fix per current session) — unaffected by the refactor (they call `macro_agent_node` / `company_agent_node`, not the service directly).

## Suggested branch
`refactor/extract-tavily-service`

## Suggested priority
- **HIGH** if #12 RiskAgent needs Tavily data (likely — risk flags need news context). Risk should consume `TavilyService` directly, not duplicate the macro/company pattern a third time.
- **MEDIUM** if #12 doesn't need Tavily. Still valuable — deduplicates macro/company, removes the source-string bug class — but not blocking.

## Coordinated issues
- **#61 (macro audit)**: fix the `source="tavily"` mislabel + backport the parse-block guard. Best to do #61 *before* this refactor: #61 fixes the existing pattern in-place; this refactor extracts the pattern. If we do this refactor first, #61 becomes "delete the lines we just wrote and call the service" — wastes the #61 work.
- **#09 CompanyAgent**: ships as-is with the FATAL guard inline. Refactor follows.

## Found by
Session 2026-08-12, issue #09 work, just before session wrap. Prady asked the architectural question unprompted — Socratic protocol: surface real tensions, then ask the user to defend the abstraction shape rather than default to "yes, file it."

pradyumna-001:
## What this PR does

- **Adds `app/services/tavily.py`** — a domain-scoped service module owning:
  - `TavilyError` hierarchy: `TavilyConfigError` (missing key — raised at construction), `TavilyRequestError` (HTTP, doc-only in v1), `TavilyParseError` (parse, doc-only in v1).
  - `TavilyArticle` frozen dataclass with `title: str | None`, `content: str | None`, `url: str`.
  - `TavilyResult` frozen dataclass with `articles: list[TavilyArticle]`, `error: DataFlag | None`.
  - `TavilyService.__init__(api_key: str, timeout: float = 10.0)` — raises `TavilyConfigError` on empty key.
  - `TavilyService.search(query: str) -> TavilyResult` — async HTTP via `httpx.AsyncClient`, parses results to typed articles, FATAL if any article has `title is None or content is None`, returns a WARNING `DataFlag` on `httpx` raise. Does NOT raise across the boundary — agents mutate state locally.
- **Rewrites `app/agents/macro.py`** — drops the inline `httpx.AsyncClient` flow + parse-block guard. Catches `TavilyConfigError` once at the node boundary (FATAL + return). Uses `result.error` (append + return) instead of inline try/except.
- **Rewrites `app/agents/company.py`** — same shape as macro. Fixes a real bug found by reading: `DataFlag(sources="tavily", ...)` (typo — would have raised `TypeError` on the `TavilyConfigError` catch path, crashing the node harder than the original failure on a real missing-key scenario).
- **Adds `tests/unit/test_tavily_service.py`** — 9 tests covering: happy path, empty `{"results": []}` (no flag), HTTP failure (WARNING), null title/content (FATAL with query in message via `!r`), partial-null articles (one valid + one invalid → FATAL), missing key raises `TavilyConfigError`, explicit key does NOT read `settings`, explicit timeout override honored.
- **Rewrites `tests/unit/test_macro_agent.py`** (5 tests) and `tests/unit/test_company_agent.py` (5 tests) — mocks now patch at the `TavilyService` boundary (`app.agents.macro.TavilyService` → `MagicMock` whose `.search` is `AsyncMock`). New `test_macro_agent_tavily_config_error` and `test_company_agent_tavily_service_error` exercise the construction-catch path the old tests never hit.
- **Adds commit `12ecb30` — fix(macro): correct `source=nvidia_nim` for LLM-None flag (#61 Part 1)** — the macro audit's mislabel was `source="tavily"` on the LLM-None branch; should have been the failing upstream (`nvidia_nim`).
- **Backports FATAL-parse rule to macro as a behavior change** (#61 Part 2 — intentionally NOT an in-place fix, superseded by the service extraction): before, Tavily returning `{"results": [{"title": null, ...}]}` silently set `macro_context=None` with no flag; after, a FATAL flag is emitted. Operators will now page on schema breaks where before they silently degraded — this is the #61 bug being fixed.

## What this PR does NOT do

- **No retry policy.** `TavilyRequestError` / `TavilyParseError` classes are kept in the hierarchy for future retry-policy discrimination but are documentation-only in v1 — `search()` builds `DataFlag` inline rather than instantiating them.
- **No rate limiting, no circuit breaker, no caching on `TavilyService`.** Out of scope for the dedup extraction.
- **`uv.lock` not staged** — carries OpenAI/httpx dep changes that belong to a separate standing decision (kept out of #09's commit scope per journal `16.md`; not introduced into #62 either).
- **No shared `ServiceError` base.** Domain-scoped like `PipelineError` per existing convention. YAGNI until a second service exists.

## How to verify

```bash
# All three gates green on the branch:
uv run ruff check .                    # full repo
uv run mypy app/services/tavily.py app/agents/macro.py app/agents/company.py
uv run pytest tests/ -v                # 35 passed, 12 skipped (DB/container-gated)

# Specifically the new + rewritten tests:
uv run pytest tests/unit/test_tavily_service.py tests/unit/test_macro_agent.py tests/unit/test_company_agent.py tests/unit/test_company_agent_stub.py tests/unit/test_macro_agent_stub.py -v
```

Implementation design doc: `.opencode/followups/extract-tavily-service.md` (110 lines, implemented as proposed with two divergences — see Decisions).

## Decisions worth flagging

- **Service boundary = `app/services/tavily.py`** (Option A placement). Matches the existing `PipelineError`-in-`app/api/errors.py` convention. Test for "does this class belong in utils/" — would a consumer of this class NOT already know about its producer? `TavilyConfigError`'s consumers all already import `app.services.tavily` → co-located.

- **Missing-key handling: raise at construction, catch once per pipeline run (Option C).** `TavilyService.__init__` raises `TavilyConfigError`; the agent node catches once, emits FATAL, sets empty state, returns. No per-agent duplication; no silent empty-result-on-construction. The failure is loud at the boundary.

- **HTTP and parse failures: returned as `TavilyResult.error: DataFlag | None`, never raised across the boundary.** Agents must mutate state (flag + set-empty + timestamp); the return-based contract keeps state mutation local to the agent. Raising would unwind through the agent without it touching `state["flags"]`.

- **Construction per-call inside the node function (Option B for placement), NOT module-level.** Module-level would import-time-crash when `TAVILY_API_KEY` is unset — breaking tests, CI (no `.env`), and hot reload, and re-introducing the `with patch("app.core.config.settings.TAVILY_API_KEY", "dummy"):` boilerplate we just eliminated.

- **Empty `{"results": []}` is VALID — no flag, `TavilyResult(articles=[], error=None)`.** Distinguishes "no matches" (Tavily honored the contract — operator not paged on slow news days) from "schema broke" (null fields in returned articles → FATAL).

- **FATAL message includes the query string via `!r`:** `f"Tavily returned articles with null title/content for query {query!r} - schema may have changed"`. The query is universally available at the service layer; ticker is agent-specific so it stays out. `!r` makes whitespace/hidden chars visible to the operator for manual re-run.

- **Layered test mocking:** agent tests mock at the service boundary (Option A — black-box); service tests mock at the `httpx` boundary (Option B — behavior-box). Each layer catches bugs the other can't see.

- **Four mock-fraud bugs found and fixed during the refactor** (all caught by reading or by the new service-layer tests, not by the existing agent-layer tests):
  1. `timenout=...` kwarg typo — invisible because mocks accept any kwarg.
  2. `all(... and ...)` vs `any(... or ...)` confusion on the FATAL check — would have mis-evaluated on real data.
  3. `AsyncMock(return_value=None)` set as `__aexit__.return_value` — silently swallowed exceptions because the mock is truthy (Python's `async with` suppresses on truthy `__aexit__` return). Led to `UnboundLocalError` on `data` because the `ConnectError` was swallowed and execution fell through. Fix: use literal `False` or remove the override entirely.
  4. `DataFlag(sources="tavily", ...)` typo in `company.py` catch path — would have crashed the very `TavilyConfigError` catch path with `TypeError`. Caught by reading, not by any existing test.

- **#61 Part 2 (parse-block guard backport to `macro.py` in-place) intentionally NOT done** — the #62 service extraction removes the need (service owns the guard now). Divergence from the follow-up doc's "do #61 in-place first, then extract" ordering; for Part 1 we respected the ordering, for Part 2 the rationale for not-duplicate-then-delete held.

- **Follow-up: audit codebase for the `AsyncMock(...) __aexit__.return_value = AsyncMock(...)` exception-suppression anti-pattern.** Mocked tests can ship silent exception-vacuums because mocks don't honor Python's truthy/falsy semantics on returned values the way real context managers do.

Closes #62
Closes #61

### DIFF
--- a/app/agents/macro.py
+++ b/app/agents/macro.py
@@ -1,78 +1,70 @@
 from datetime import datetime, UTC
-import logging
 
-import httpx
+import logging
 
 from app.core.config import settings
 from app.graph.state import AgentState, MacroOutput
 from app.prompts.macro import MACRO_PROMPTS
 from app.services.llm import summarize
+from app.services.tavily import TavilyService, TavilyConfigError
 from app.utils.flags import DataFlag, Severity
+
 
 logger = logging.getLogger(__name__)
 
 
 async def macro_agent_node(state: AgentState) -> AgentState:
-    """Pull macro-news via Tavily, build a MacroOutput, and update
-    the AgentState in-place. Failures append a DataFlag and return
-    with macro_context=None — never raises.
+    """Pull macro-news via TavilyService, build a MacroOutput, and update
+    AgentState in-place. Failures append a DataFlag and returns
+    with macro_context=None - never raises.
     """
     logger.info(
         "macro_agent_start",
         extra={
             "pipeline_run_id": state["pipeline_run_id"],
             "morning_note_id": state["morning_note_id"],
-            "manager_id": state["manager_id"],
-        },
+            "manager_id": state["manager_id"]
+        }
     )
 
-    tavily_key = settings.TAVILY_API_KEY
-    if not tavily_key:
+    try:
+        tavily = TavilyService(api_key=settings.TAVILY_API_KEY)
+    except TavilyConfigError as exc:
         state["flags"].append(
             DataFlag(
                 source="tavily",
                 severity=Severity.FATAL,
-                message="TAVILY_API_KEY missing",
+                message=str(exc)
             )
         )
         state["macro_context"] = None
         state["data_freshness"]["macro"] = datetime.now(UTC)
         return state
 
-    payload = {"query": "macro news Brazil", "max_results": 10,
-               "include_domains": ["bcb.gov.br", "ibge.gov.br", "br.reuters.com", "bloomberg.com.br"]}
-    try:
-        async with httpx.AsyncClient(timeout=10.0) as client:
-            resp = await client.post(
-                "https://api.tavily.com/search",
-                headers={"Authorization": f"Bearer {tavily_key}"},
-                json=payload,
-            )
-            resp.raise_for_status()
-            data = resp.json()
-    except Exception as exc:
-        state["flags"].append(
-            DataFlag(
-                source="tavily",
-                severity=Severity.WARNING,
-                message=f"Tavily request error: {exc}",
-            )
-        )
+    result = await tavily.search(
+        query="macro news Brazil",
+        include_domains=[
+            "bcb.gov.br", 
+            "ibge.gov.br", 
+            "br.reuters.com", 
+            "bloomberg.com.br"
+        ],
+        max_results=10
+    )
+    if result.error:
+        state["flags"].append(result.error)
         state["macro_context"] = None
         state["data_freshness"]["macro"] = datetime.now(UTC)
         return state
 
     macro_output: MacroOutput | None = None
-    if data and data.get("results"):
-        top = data["results"][0]
-        raw_text = (
-            f"{top.get('title', '')}\n\n"
-            f"{top.get('content', '')}\n{top.get('url', '')}"
-        )
+    if result.articles:
+        top = result.articles[0]
+        raw_text = f"{top.title or ''}\n\n{top.content or ''}\n{top.url or ''}"
 
         summary = await summarize(
             system=MACRO_PROMPTS.system,
-            user=MACRO_PROMPTS.build_user_prompt(raw_text=raw_text),
+            user=MACRO_PROMPTS.build_user_prompt(raw_text=raw_text)
         )
 
         if summary is None:
@@ -80,16 +72,16 @@ async def macro_agent_node(state: AgentState) -> AgentState:
                 DataFlag(
                     source="nvidia_nim",
                     severity=Severity.WARNING,
-                    message="Summarization returned None; using raw content.",
+                    message="Summarization returned None; using raw content"
                 )
             )
-            summary = top.get("content", "")
+            summary = top.content or ""
 
         macro_output = MacroOutput(
-            headline=str(top.get("title", "")),
+            headline=top.title or "",
             summary=summary,
-            sources=[str(top.get("url", ""))] if top.get("url") else [],
-            fetched_at=datetime.now(UTC).isoformat(),
+            sources=[top.url] if top.url else [],
+            fetched_at=datetime.now(UTC).isoformat()
         )
 
     state["macro_context"] = macro_output
--- a/app/agents/macro.py
+++ b/app/agents/macro.py
@@ -78,7 +78,7 @@ async def macro_agent_node(state: AgentState) -> AgentState:
         if summary is None:
             state["flags"].append(
                 DataFlag(
-                    source="tavily",
+                    source="nvidia_nim",
                     severity=Severity.WARNING,
                     message="Summarization returned None; using raw content.",
                 )
--- a/app/services/tavily.py
+++ b/app/services/tavily.py
@@ -1,0 +1,103 @@
+from __future__ import annotations
+
+import httpx
+from dataclasses import dataclass
+
+from app.core.config import settings
+from app.utils.flags import DataFlag, Severity
+
+
+class TavilyError(Exception):
+    """Base for all TavilyService failures. Callers can `except TavilyError:` as catch-all."""
+
+
+class TavilyConfigError(TavilyError):
+    """Raised when the service cannot be constructed (e.g. missing API key)"""
+
+
+class TavilyRequestError(TavilyError):
+    """Internal: HTTP/network failure. Surfaced via TavilyResult.error, never raised across the boundary"""
+
+
+class TavilyParseError(TavilyError):
+    """Internal: bad response shape. Surfaced via TavilyResult.error, never raised across the boundary."""
+
+
+@dataclass(frozen=True)
+class TavilyArticle:
+    title: str | None
+    content: str | None
+    url: str | None
+
+
+@dataclass(frozen=True)
+class TavilyResult:
+    articles: list[TavilyArticle]
+    error: DataFlag | None
+
+
+class TavilyService:
+    def __init__(self, api_key: str | None = None, timeout: float = 10.0) -> None:
+        key = api_key if api_key is not None else settings.TAVILY_API_KEY
+        if not key:
+            raise TavilyConfigError("TAVILY_API_KEY missing or empty")
+        self.api_key = key
+        self.timeout = timeout
+
+    async def search(
+            self,
+            query: str,
+            include_domains: list[str],
+            max_results: int = 10
+    ) -> TavilyResult:
+        payload = {
+            "query": query,
+            "max_results": max_results,
+            "include_domains": include_domains
+        }
+        try:
+            async with httpx.AsyncClient(timeout=self.timeout) as client:
+                resp = await client.post(
+                    "https://api.tavily.com/search",
+                    headers={"Authorization": f"Bearer {self.api_key}"},
+                    json=payload
+                )
+                resp.raise_for_status()
+                data = resp.json()
+        except Exception as exc:
+            return TavilyResult(
+                articles=[],
+                error=DataFlag(
+                    source="tavily",
+                    severity=Severity.WARNING,
+                    message=f"Tavily request error: {exc}"
+                )
+            )
+
+        raw_articles = data.get("results") or []
+        articles: list[TavilyArticle] = []
+        for raw in raw_articles:
+            if not isinstance(raw, dict):
+                continue
+            article = TavilyArticle(
+                title=raw.get("title"),
+                content=raw.get("content"),
+                url=raw.get("url")
+            )
+            articles.append(article)
+
+        if articles and any(a.title is None or a.content is None for a in articles):
+            return TavilyResult(
+                articles=[],
+                error=DataFlag(
+                    source="tavily",
+                    severity=Severity.FATAL,
+                    message=(
+                        f"Tavily returned articles with null title/content for "
+                        f"query {query!r} - schema may have changed"
+                    )
+                )
+            )
+
+        return TavilyResult(articles=articles, error=None)
+    
--- a/app/agents/company.py
+++ b/app/agents/company.py
@@ -1,12 +1,12 @@
 from datetime import datetime, UTC
 
 import logging
-import httpx
 
 from app.core.config import settings
 from app.graph.state import AgentState, CompanyEvent
 from app.prompts.company import COMPANY_PROMPTS
 from app.services.llm import summarize
+from app.services.tavily import TavilyService, TavilyConfigError
 from app.utils.flags import DataFlag, Severity
 
 
@@ -14,7 +14,7 @@ logger = logging.getLogger(__name__)
 
 
 async def company_agent_node(state: AgentState) -> AgentState:
-    """Pull company news via Tavily for the ticker in state, extract
+    """Pull company news via TavilyService for ticker in state, extract
     CompanyEvents via NVIDIA NIM, and update AgentState in-place.
     Failures append a DataFlag and return - never raises.
     """
@@ -27,67 +27,40 @@ async def company_agent_node(state: AgentState) -> AgentState:
         }
     )
 
-
-    tavily_key = settings.TAVILY_API_KEY
-    if not tavily_key:
+    try:
+        tavily = TavilyService(api_key=settings.TAVILY_API_KEY)
+    except TavilyConfigError as exc:
         state["flags"].append(
             DataFlag(
                 source="tavily",
                 severity=Severity.FATAL,
-                message="TAVILY_API_KEY missing"
+                message=str(exc)
             )
         )
         state["company_events"] = []
         state["data_freshness"]["company"] = datetime.now(UTC)
         return state
 
-
     ticker = state["company_ticker"]
-    payload = {
-        "query": f"{ticker} news Brazil",
-        "max_results": 10,
-        "include_domains": [
+    result = await tavily.search(
+        query=f"{ticker} news Brazil",
+        include_domains=[
             "cvm.gov.br",
             "infomoney.com.br",
             "globo.com/valor-economico"
-        ]
-    }
-    try:
-        async with httpx.AsyncClient(timeout=10.0) as client:
-            resp = await client.post(
-                "https://api.tavily.com/search",
-                headers={"Authorization": f"Bearer {tavily_key}"},
-                json=payload
-            )
-            resp.raise_for_status()
-            data = resp.json()
-    except Exception as exc:
-        state["flags"].append(
-            DataFlag(
-                source="tavily",
-                severity=Severity.WARNING,
-                message=f"Tavily request error: {exc}"
-            )
-        )
+        ],
+        max_results=10
+    )
+    if result.error:
+        state["flags"].append(result.error)
         state["company_events"] = []
         state["data_freshness"]["company"] = datetime.now(UTC)
         return state
 
-
     company_events: list[CompanyEvent] = []
-    try:
-        if not data or not data.get("results"):
-            raise ValueError("Tavily returned no results field")
-        top = data["results"][0]
-        if not (top.get("title") and top.get("content")):
-            raise ValueError(
-                f"Tavily returned results with null title/content for "
-                f"{ticker} - schema may have changed"
-            )
-        raw_text = (
-            f"{top.get('title', '')}\n\n"
-            f"{top.get('content', '')}\n{top.get('url', '')}"
-        )
+    if result.articles:
+        top = result.articles[0]
+        raw_text = f"{top.title or ''}\n\n{top.content or ''}\n{top.url or ''}"
 
         summary = await summarize(
             system=COMPANY_PROMPTS.system,
@@ -102,27 +75,16 @@ async def company_agent_node(state: AgentState) -> AgentState:
                     message="Summarization returned None; using raw content"
                 )
             )
-            summary = top.get("content", "")
+            summary = top.content or ""
 
         company_events.append(
             CompanyEvent(
-                title=str(top.get("title", "")),
+                title=top.title or "",
                 date="",
-                source=str(top.get("url", "")),
+                source=top.url or "",
                 summary=summary
             )
         )
-    except (ValueError, KeyError, IndexError) as exc:
-        state["flags"].append(
-            DataFlag(
-                source="tavily",
-                severity=Severity.FATAL,
-                message=f"parse-block failed for {state['company_ticker']}: {exc}"
-            )
-        )
-        state["company_events"] = []
-        state["data_freshness"]["company"] = datetime.now(UTC)
-        return state
 
     state["company_events"] = company_events
     state["data_freshness"]["company"] = datetime.now(UTC)
--- a/tests/unit/test_company_agent.py
+++ b/tests/unit/test_company_agent.py
@@ -1,105 +1,116 @@
+import logging
+from datetime import datetime
+from unittest.mock import patch, AsyncMock
+
 import pytest
 
-from unittest.mock import AsyncMock, patch, MagicMock
-
 from app.agents.company import company_agent_node
-from app.graph.state import create_initial_state, Severity
+from app.graph.state import create_initial_state, DataFlag
+from app.services.tavily import TavilyArticle, TavilyResult
+from app.utils.flags import Severity
+
+
+def make_state():
+    return create_initial_state(
+        manager_id=1,
+        company_ticker="PETR4",
+        pipeline_run_id="run",
+        morning_note_id="note",
+    )
+
+
+def _tavily_result(articles=None, error=None):
+    return TavilyResult(articles=articles or [], error=error)
 
 
 @pytest.mark.asyncio
 async def test_company_agent_no_key():
-    with patch("app.agents.company.settings.TAVILY_API_KEY", ""):
-        state = create_initial_state(
-            manager_id=1, company_ticker="PETR4",
-            pipeline_run_id="run", morning_note_id="note",
-        )
+    from app.services.tavily import TavilyConfigError
+    with patch("app.agents.company.TavilyService", side_effect=TavilyConfigError("TAVILY_API_KEY missing or empty")):
+        state = make_state()
         result = await company_agent_node(state)
 
     assert result["company_events"] == []
-    assert any(
-        f.source == "tavily" and f.severity == Severity.FATAL
-        for f in result["flags"]
-    )
-    from datetime import datetime
+    flag = result["flags"][-1]
+    assert isinstance(flag, DataFlag)
+    assert flag.source == "tavily"
+    assert flag.severity == Severity.FATAL
     assert isinstance(result["data_freshness"]["company"], datetime)
 
 
 @pytest.mark.asyncio
 async def test_company_agent_llm_none():
-    tavily_json = {
-        "results": [
-            {
-                "title": "PETR4 earnings Q3",
-                "content": "Petrobras reported record earnings...",
-                "url": "https://infomoney.com.br/petr4",
-            }
-        ]
-    }
+    article = TavilyArticle(
+        title="PETR4 earnings Q3",
+        content="Petrobras reported record earnings...",
+        url="https://infomoney.com.br/petr4",
+    )
+    result = _tavily_result(articles=[article])
 
-    mock_client = AsyncMock()
-    mock_resp = MagicMock()
-    mock_resp.json.return_value = tavily_json
-    mock_resp.raise_for_status = MagicMock()
-    mock_client.post.return_value = mock_resp
-    mock_client.__aenter__.return_value = mock_client
+    with patch("app.agents.company.TavilyService") as MockService:
+        MockService.return_value.search = AsyncMock(return_value=result)
+        with patch("app.agents.company.summarize", new=AsyncMock(return_value=None)):
+            state = make_state()
+            res = await company_agent_node(state)
 
-    async def summarize_returns_none(system, user):
-        return None
-
-    with patch("app.agents.company.settings.TAVILY_API_KEY", "dummy"):
-        with patch("app.agents.company.httpx.AsyncClient", return_value=mock_client):
-            with patch("app.agents.company.summarize", side_effect=summarize_returns_none):
-                state = create_initial_state(
-                    manager_id=1, company_ticker="PETR4",
-                    pipeline_run_id="run", morning_note_id="note",
-                )
-                result = await company_agent_node(state)
-
-    events = result["company_events"]
+    events = res["company_events"]
     assert len(events) == 1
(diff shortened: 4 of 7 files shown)


app/services/tavily.py
```diff
@@ -0,0 +1,103 @@
...
+class TavilyResult:
+    articles: list[TavilyArticle]
+    error: DataFlag | None
+
+
+class TavilyService:
+    def __init__(self, api_key: str | None = None, timeout: float = 10.0) -> None:
+        key = api_key if api_key is not None else settings.TAVILY_API_KEY
+        if not key:
+            raise TavilyConfigError("TAVILY_API_KEY missing or empty")
```
pradyumna-001, line 43: Q1 (architecture) — raise vs return across the boundary.

`TavilyConfigError` is raised at construction (line 43), while `TavilyRequestError` and `TavilyParseError` are surfaced as `TavilyResult.error` and never raised across the boundary. That's an intentional asymmetry — but a teammate reviewing this would want to hear the principle in your words, not just see it in the diff.

Walk me through: what property of "missing key" makes raise-at-construction the right shape, and what property of "HTTP failed" or "payload was unparseable" makes return-the-flag the right shape? Specifically — is there a state-mutation reason the agent must perform on the request/parse paths that it does NOT need to perform on the missing-key path, and does that reason drive the return-vs-raise split?

app/services/tavily.py
```diff
@@ -0,0 +1,103 @@
...
+
+class TavilyConfigError(TavilyError):
+    """Raised when the service cannot be constructed (e.g. missing API key)"""
+
+
+class TavilyRequestError(TavilyError):
+    """Internal: HTTP/network failure. Surfaced via TavilyResult.error, never raised across the boundary"""
+
+
+class TavilyParseError(TavilyError):
```
pradyumna-001, line 22: Q2 (architecture) — YAGNI vs speculative scaffolding.

`TavilyRequestError` and `TavilyParseError` are defined but never instantiated in `search()` — the code builds `DataFlag` inline (lines 68–74 and 89–100) and returns it directly, so the two subclasses are documentation-only in v1. Their docstrings even say "never raised across the boundary."

The usual YAGNI instinct is to delete classes that carry no behavior. You kept them. Defend that — what future discriminator (retry policy? alert routing? metrics tagging?) justifies keeping the classes around *now*, and at what point would you decide they should be deleted instead of kept as placeholders?

app/services/tavily.py
```diff
@@ -0,0 +1,103 @@
...
+            if not isinstance(raw, dict):
+                continue
+            article = TavilyArticle(
+                title=raw.get("title"),
+                content=raw.get("content"),
+                url=raw.get("url")
+            )
+            articles.append(article)
+
+        if articles and any(a.title is None or a.content is None for a in articles):
```
pradyumna-001, line 89: Q3 (architecture) — the FATAL predicate semantics.

Line 89: `if articles and any(a.title is None or a.content is None for a in articles)` → FATAL, and you set `articles=[]` in the returned `TavilyResult` (line 91). Two layers of judgment are stacked here that a teammate would want unpacked:

1. Why `any(... or ...)` — i.e., *any* article in the batch with *either* null field trips FATAL for the whole batch — rather than per-article filtering the valid ones and only flagging if *all* are bad?
2. Why does FATAL discard the entire `articles` list (line 91 sets `articles=[]`) instead of returning the partial-but-valid subset alongside the flag?

The operator-paging angle matters here: a slow-news day returns `{"results": []}` (no flag, line 102), but a day where Tavily returns 5 articles and 1 has a null title silently zero-powers the other 4 and pages you. Is that the right policy, or did you consider partial-batch survival?

app/agents/macro.py
```diff
@@ -1,97 +1,89 @@
...
-            "manager_id": state["manager_id"],
-        },
+            "manager_id": state["manager_id"]
+        }
     )
 
-    tavily_key = settings.TAVILY_API_KEY
-    if not tavily_key:
+    try:
+        tavily = TavilyService(api_key=settings.TAVILY_API_KEY)
```
pradyumna-001, line 31: Q4 (architecture) — construction per-call, not module-level.

Line 31: `tavily = TavilyService(api_key=settings.TAVILY_API_KEY)` is constructed *inside* `macro_agent_node` on every invocation, not once at module load. `TavilyConfigError` therefore only fires when the node runs, not at import time — and tests like `test_macro_agent_no_key` exercise the catch path by patching `settings.TAVILY_API_KEY` rather than via module-level construction.

The case against module-level is loud (import-time crash, CI breakage, the old `with patch("settings.TAVILY_API_KEY")` boilerplate). But the case *for* per-call is subtler — there's a real cost: a new service instance every pipeline run, reading `settings` each call, no opportunity to inject a configured-long-lived client or connection pooling. Walk me through the property that made "per-call, read settings each time" the right call, and where you'd expect that decision to age poorly (e.g., when does per-call construction start to hurt — at what workload shape would you revisit)?

app/services/tavily.py
```diff
@@ -0,0 +1,103 @@
+from __future__ import annotations
+
+import httpx
+from dataclasses import dataclass
+
+from app.core.config import settings
+from app.utils.flags import DataFlag, Severity
+
+
+class TavilyError(Exception):
```
pradyumna-001, line 10: Q5 (architecture) — domain-scoped base, no shared `ServiceError`.

`TavilyError` lives in `app/services/tavily.py` and inherits straight from `Exception` (line 10). There's no `app/services/errors.py` with a shared `ServiceError` base, and no `app/utils/service_error.py` either. `PipelineError` over in `app/api/errors.py` follows the same pattern — domain-scoped, no shared ancestor.

A teammate who's seen a few codebases would expect the temptation to be: "I'm about to add a second service (TavilyService → next will be LLMService, or AlphaVantageService), I should drop a `ServiceError` base into utils now and future-proof."
You didn't. Defend the YAGNI call — what's the signal that would tell you it's time to introduce the shared base (a third service? a shared retry decorator that needs to catch across services?), and is there a risk that waiting means TavilyError and (future) OtherServiceError drift apart in shape (different __init__ arities, different severity enums) and make later unification painful rather than trivial?

pradyumna-001:
Peer review (self-review pre-merge) — 5 architectural questions posted as inline comments on the diff. Tagged Q1–Q5 for traceability. I'll defend each in a reply on its own thread.

- **Q1** `tavily.py:43` — raise `TavilyConfigError` vs return `TavilyResult.error` (the asymmetry).
- **Q2** `tavily.py:22` — keep doc-only `TavilyRequestError`/`TavilyParseError` (YAGNI challenge).
- **Q3** `tavily.py:89` — `any(... or ...)` FATAL predicate + full-batch discard vs partial survival.
- **Q4** `macro.py:31` — construction per-call vs module-level (cost of per-call).
- **Q5** `tavily.py:10` — domain-scoped base, no shared `ServiceError`.

Defenses incoming on each thread.

app/services/tavily.py
pradyumna-001: The raise-vs-return split tracks future-recoverability of the failure, not the agent's response shape. 

Missing key is unrecoverable by any caller decision. So raise says something like "this failure is unrecoverable by you, the caller; only the operator can fix it". On the other hand HTTP/parse returns because a future caller could retry, fall back to a different source or even use partial data. Raise would unwind immediately and eliminate that option.

Raising TavilyRequestError across the boundary would mean the agent knows about the retry policy, source priorities and degradation rules. Returning the flag keeps the decision with the agent where it belongs.

app/services/tavily.py
pradyumna-001: The classes are declared for future retry policy. TavilyRequestError and TavilyParseError have opposite retry semantics - one is retry-worthy and the other isn't. I'm keeping them because the hierarchy skeleton encodes a discrimination that's hard to re-discover later. 

They can be used as documentation-like code to a future implementation. A future developer that reads them can know exactly what they're for and can use them. The principle is: the developer will be aware of the idea behind those classes just by reading the docstring instead of thinking of a new solution that actually has been thought before.

If a second service ships and still no retry decorator consumes these, that's the signal that the skeleton was speculative - delete them. If a retry decorator ships and only discriminates on Severity, not on exception type - same conclusion, delete the classes and let the decorator branch on severity.

pradyumna-001:
Q1 ✅ Solved — defense posted as reply on the `tavily.py:43` thread. Principle: raise-vs-return split tracks future-recoverability of the failure mode, not the agent's response shape.

pradyumna-001:
Q2 ✅ Solved — defense updated on the `tavily.py:22` thread. Principle: keep the doc-only exception subclasses because they encode the transient-vs-persistent retry discrimination that a future retry decorator will need; delete only if a second service ships without a retry consumer or a decorator branches on Severity instead of exception type.

app/services/tavily.py
pradyumna-001: Partial-survival is right only if the agent's contract lets it continue past a FATAL flag, so returning a valid subset alongside a FATAL flag is dead code - the agent discards it before reaching the summarizer.

Empty {results: []} honors the contract (operator does nothing); 1-null-in-5 breaks the contract (schema drifted, operator must investigate). FATAL on the second pages the operator when action is needed; no-flag on the first avoids crying wolf on slow news days.

To actually use the subset, the agent contract has to change to "error -> flag-and-continue-with-partial", which is a bigger change than this PR is scoped to make. If I returned articles=[valid] + FATAL, the agent bails on if result.error and the valid list is silently discarded. So, full-discard is honest about what the contract does.

pradyumna-001:
Q3 ✅ Solved — defense posted on the `tavily.py:89` thread. Principle: partial-survival doesn't compose with the agent's existing 'error → bail' contract (macro.py:49-53), so returning a valid subset alongside a FATAL flag is dead code; full-discard is honest about what the contract does. Empty-results honors the contract (no flag), 1-null-in-5 breaks it (FATAL pages the operator when action is needed).

app/agents/macro.py
pradyumna-001: Per-call construction is right when construction is cheap and what sharing would buy is gated by a separate decision. It's cost is a dict lookup + two assignments + one branch - negligible against await tavily.search() which performs an HTTP round-trip.

A module-level service would only pay off if it held a long-lived pooled httpx.AsyncClient. But search() already opens a fresh client per call, so even this way we could have a pool connection without a second change.

If bached/looped calls withing a single pipeline run become hot enought that TCP/TLS handshake overhead matters or a second serivce lands that wants the same pool, then we should revisit per-call construction. Today's per-call choice is correct because of the cost-of-per-call equals to cost-of-per-call-within-module-level until the pooling decision is made.

pradyumna-001:
Q4 ✅ Solved — defense posted on the `macro.py:31` thread. Principle: per-call construction is right when construction is cheap and what sharing would buy is gated by a separate decision. Cost is dict-lookup + two assignments + one branch (no socket, no `httpx.AsyncClient` opened at init — client is fresh inside `search()`). Revisit when batched calls within a pipeline run get hot enough that TCP/TLS handshake matters, or when a second service wants the same pool.

app/services/tavily.py
pradyumna-001: A shared abstraction earns its keep when a consumer exists that needs to catch across services without knowing which. Today, each agent catches TavilyConfigError specifically; neither catches a hypothetical ServiceError. So, the shared base has zero consumers.

PipelineError or a DataFlag each have one clear reason for existing. A ServiceError base would carries empty marker bases with no shared behavior that are the dead-weight abstraction - they exist to feel future-proof, not to do anythin.

If you wait, TavilyError and future OtherServiceError might diverge in __init__ arity and unification becomes painful. But the shared base can always be introduced when the second service arrives, because introducing it only requires TavilyError(ServiceError), OtherServiceError(ServiceError) - simple mechanical re-parenting, no consumer code changes.

The signal that would tell us it's time to introduce the shared base is when a consumer emerges that need cross-service catching. At that point the base has real consumer and earns its keep.

Issue #69, opened 2026-08-13
Chore: canonicalize DataFlag import in test files (use app.utils.flags, not app.graph.state re-export)


pradyumna-001:
Filed during #10 QuantAgent work (session 2026-08-13).

## Context

`DataFlag` and `Severity` are defined in `app/utils/flags.py`. Several test files import `DataFlag` (and in one case `Severity`) from `app.graph.state` instead — which works because `state.py:4` re-exports them via `from app.utils.flags import DataFlag, Severity`, but the re-export path is non-canonical and confuses readers ("why does state export a Flag class?").

Files importing from `app.graph.state`:

- `tests/unit/test_macro_agent.py:8` — `from app.graph.state import create_initial_state, DataFlag`
- `tests/unit/test_company_agent.py:8` — `from app.graph.state import create_initial_state, DataFlag`
- `tests/unit/test_company_agent_stub.py:9` — `from app.graph.state import create_initial_state, DataFlag`
- `tests/unit/test_quant_agent.py:8` — `from app.graph.state import create_initial_state, DataFlag` (#10)
- `tests/unit/test_macro_agent_stub.py:4` — `from app.graph.state import create_initial_state, Severity`

## Canonical form

Split into two imports in each file:

```python
from app.graph.state import create_initial_state
from app.utils.flags import DataFlag, Severity
```

(Or keep the state import for `create_initial_state` and add a separate utils import for `DataFlag` / `Severity` — pick the shape that reads cleanest.)

## Functional impact

Zero. Both import paths resolve to the same class object; tests pass under either form. The change is purely for reader clarity — imports should point at the canonical home, not a re-export.

## Scope

~5 test files, mechanical rename. No production code touched. No test logic changes.

## Suggested branch

`chore/canonicalize-dataflag-imports`

## Suggested priority

Low — pure cosmetic cleanup. Pick up during a cleanup pass alongside #66 (dead `freshness.py` audit) and #67 (delete doc-only Tavily subclasses). The three together would make a natural cleanup PR.

## Not blocking #10

#10 ships with the non-canonical import to match existing test-file convention. This issue tracks the canonicalization across all test files.

Issue #68, opened 2026-08-13
Feature: add market_time to QuantOutput — blocks RiskAgent #12


pradyumna-001:
Filed during #10 QuantAgent work (session 2026-08-13).

## Context

`QuantOutput` (`app/graph/state.py:21-27`) currently has `fetched_at: str` — when QuantAgent fetched the data. It does NOT carry `regularMarketTime` (yfinance's source-side timestamp for when the market last traded this ticker).

During #10, two candidate semantics for `state["data_freshness"]["quant"]` were considered:
- **α (shipped in #10)**: `datetime.now(UTC)` — fetch time. Matches macro/company agents' pattern.
- **β1 (deferred to this issue)**: store `regularMarketTime`-derived `market_time: datetime` in `QuantOutput`, plus `state["data_freshness"]["quant"]` switches to data-time.

α shipped because no downstream consumer needs data-time today (Q5 consumer-test applied five times running). β1 is deferred until a real consumer arrives.

## Why β1 is blocking RiskAgent #12

`finagent-CLAUDE.md:61` says "QuantAgent must verify `data_freshness` before calculating any metric." RiskAgent #12 also reads `quant_metrics` when forming its adversarial risk assessment — and a real question surfaced during #10's design: **does RiskAgent need `market_time` (data-time) to time-correlate news/events against the quant data's cut-off, or is `fetched_at` (fetch-time) enough?**

If RiskAgent needs data-time, β1 has to land *before* RiskAgent can be designed. If RiskAgent doesn't need it, β1 stays deferred and this issue closes.

So this issue is a **prerequisite decision** for #12's design phase, not necessarily an implementation that has to merge before #12's branch starts — the decision ("does RiskAgent need `market_time`?") has to be made before #12's first chunk is authored.

## Trigger

Start of RiskAgent #12 design phase. First architectural decision: does RiskAgent consume `quant_metrics["market_time"]`? If yes → implement this issue first. If no → close this issue as not-needed.

### Schema change

`app/graph/state.py`:
```python
class QuantOutput(TypedDict, total=False):
    pl: float
    ev_ebitda: float
    p_vpa: float
    dividend_yield: float
    dev_ibov: float
    fetched_at: str
    market_time: datetime  # NEW — when the market last produced this data
```

### Service plumbing

`app/services/yfinance.py`:
- `QuoteMetrics` gains `market_time: datetime | None`.
[...]

Issue #67, opened 2026-08-13
Chore: delete doc-only TavilyRequestError + TavilyParseError from tavily.py (Q2 deletion-trigger fired)


pradyumna-001:
Filed during #10 QuantAgent work (session 2026-08-13).

## Context — the deletion-trigger from PR #64's Q2 defense

In the peer review of PR #64 (Extract TavilyService), the defense on question Q2 (keep doc-only `TavilyRequestError` + `TavilyParseError` as YAGNI placeholders for a future retry decorator) explicitly named a deletion-trigger:

> *"If a second service ships and still no retry decorator consumes these, that's the signal that the skeleton was speculative - delete them."*

Today's state:
- Second service is shipping (#10 QuantAgent → `YfinanceService`, in progress on `feature/issue-10-quant-agent`).
- No retry decorator exists in the repo.
- No consumer catches `TavilyRequestError` or `TavilyParseError` as a type — `TavilyService.search()` builds `DataFlag` inline with `severity=WARNING` (HTTP transient) and `severity=FATAL` (parse persistent). The severity enum carries the discrimination; the exception types do nothing.

**Trigger fired.** The subclasses are confirmed speculative — keeping them contradicts the discipline applied across four other decisions (#62 Q5 ServiceError, #62 Q2 the subclasses themselves, #10 Decision #1 QuoteService abstraction, #10 Decision #3b multi-horizon `dev_ibov`).

### `app/services/tavily.py`

Delete the two classes:

```python
class TavilyRequestError(TavilyError):
    """Internal: HTTP/network failure. Surfaced via TavilyResult.error, never raised across the boundary"""


class TavilyParseError(TavilyError):
    """Internal: bad response shape. Surfaced via TavilyResult.error, never raised across the boundary."""
```

Keep `TavilyError` (the catch-all base — still needed for `except TavilyError:` catch-all by callers) and `TavilyConfigError` (raised at construction time for missing key — actually used by `macro_agent_node` and `company_agent_node`).

The discrimination that lives in `DataFlag.severity` (WARNING for HTTP transient, FATAL for parse persistent) is the load-bearing signal — no behavior change in `search()` or the agents.

### `app/services/yfinance.py`

`YfinanceService` (in progress on #10 branch) ships with only:

- `YfinanceError` (catch-all base)
- `YfinanceConfigError` — **NO** (no API key, no construction failure)

[...]

Issue #66, opened 2026-08-13
Chore: delete unused app/utils/freshness.py + test_freshness.py (dead code audit)


pradyumna-001:
Filed during #10 QuantAgent work (session 2026-08-13).

## Context

`app/utils/freshness.py` exports `is_data_fresh(source, fetched_at, now, max_age=timedelta(hours=24)) -> bool`. A repo-wide grep confirms it has **zero production callers** — only `tests/unit/test_freshness.py` references it. MacroAgent and CompanyAgent stamp `datetime.now(UTC)` into `state["data_freshness"][...]` directly, without calling this utility.

QuantAgent (#10) will introduce the first real freshness *check* in the codebase, but the check compares against yfinance's source-side `regularMarketTime` (data age), not against fetch time. So the new check lives inline in `YfinanceService` (`app/services/yfinance.py`) and doesn't use `freshness.py` either.

Result: `freshness.py` remains permanently dead as far as any current or #10 consumer goes.

## Latent bug (orthogonal, but worth noting)

`freshness.py:6` uses `datetime.now()` without `tz=UTC`. The agents all stamp `datetime.now(UTC)` (aware datetimes); subtracting a naive `now` from an aware `fetched_at` raises `TypeError`. The bug is moot while the function is dead, but a future caller who revives it will hit it.

## Scope of this chore

One of two decisions should be made in this PR:

**Option A — delete.** Remove `app/utils/freshness.py` and `tests/unit/test_freshness.py`. The "QuantAgent must verify `data_freshness`" rule (CLAUDE.md:61) is enforced by `YfinanceService` directly; no generic utility needed.

**Option B — revive.** Fix the UTC bug at `:6`, generalize the signature to accept a source-side timestamp (so it can serve both fetch-latency and data-age checks), and wire it up from `YfinanceService`. Makes it a real canonical entry point for freshness checks across agents.

## Suggested branch

`chore/cleanup-dead-freshness`

## Suggested priority

Low — not blocking #10 or any active issue. Pick up during a cleanup session or when adding the second agent that needs a freshness *check* (not just a freshness *stamp*).

## Not blocking #10

#10 ships without touching `freshness.py`. The inline check in `YfinanceService` is self-contained.

PR #70, opened 2026-08-13
feat(quant): Closes #10 — QuantAgent with YfinanceService (no LLM for calculations)

### LINKED ISSUE #10: arch: design and implement unified prompt management service and schemas

pradyumna-001:
## What this PR does

- **Adds `app/services/yfinance.py`** — `YfinanceService` mirroring `TavilyService` (extracted in #62). Domain-scoped service owning:
  - `YfinanceError` — catch-all base only (no `YfinanceConfigError`, no `YfinanceRequestError`/`YfinanceParseError` subclasses; per #67's deletion-trigger discipline, no doc-only subclasses on a new service when no retry-decorator consumer exists).
  - `QuoteMetrics` frozen dataclass with `pl`, `ev_ebitda`, `p_vpa`, `dividend_yield`, `dev_ibov`, `fetched_at` — field names match `QuantOutput` exactly (per Decision #1: yfinance's `.info` key names (`trailingPE`, `enterpriseToEbitda`, `priceToBook`, `dividendYield`) are normalized at the service boundary; agents never see yfinance's internal keys).
  - `YfinanceResult` frozen dataclass with `metrics: QuoteMetrics | None` + `error: DataFlag | None`.
  - `YfinanceService.__init__(ticker, market_index="^BVSP", max_data_age_hours=24)` — raises `YfinanceError` on empty ticker (unrecoverable by caller, raise per Q1-of-PR-#64 future-recoverability axis).
  - `YfinanceService.search()` — async; uses `asyncio.to_thread(lambda: yf.Ticker(...).info)` to offload the sync yfinance call.
- **Adds `app/agents/quant.py`** — `quant_agent_node`. Catches `YfinanceError` at construction (FATAL flag + return); reads `result.error` post-`search()` (FATAL or WARNING flag + return); builds `QuantOutput` from `QuoteMetrics` fields; stamps `state["data_freshness"]["quant"] = datetime.now(UTC)` (Option α — fetch-time; β1 data-time deferred to #68 when RiskAgent #12 lands).
- **Modifies `app/graph/state.py`** — `QuantOutput` field types widened from `float` to `float | None` per the partial-success principle (a non-dividend-paying company legitimately has `dividendYield: None`; the `| None` encodes "metric not applicable" at the type level).
- **Adds `tests/unit/test_yfinance_service.py`** — 9 service-layer tests mocking at the `yfinance` module boundary (Option B per #62's layered-test discipline — behavior-box for the service). Uses `side_effect` dispatch on `yf.Ticker(...)` per-symbol (`PETR4.SA` vs `^BVSP`) so the IBOV and ticker fetch get different mock returns.
- **Adds `tests/unit/test_quant_agent.py`** — 5 agent-layer tests mocking at the `YfinanceService` boundary (Option A per #62 — black-box for the agent).
- **Adds `scripts/run_quant_agent.py`** — live-data diagnostic script. First agent that runs live without needing API keys (yfinance is unauthenticated); catches shape-mismatches in yfinance's return that the unit tests structurally cannot (see fifth mock-fraud below).
- **Adds yfinance>=1.6.0 to `pyproject.toml` + `uv.lock`** — first commit ships the dep with the service skeleton.

## What this PR does NOT do

- **Does NOT use LLM for any calculation.** CLAUDE.md:175 binding rule: "QuantAgent does NOT use LLM for calculations." This PR computes `pl`, `ev_ebitda`, `p_vpa`, `dividend_yield` directly from yfinance's `.info`, and `dev_ibov` from close-price arithmetic. LLM interpretation of results is deferred to RiskAgent #12 / EditorAgent.
- **Does NOT add `YfinanceConfigError`/`YfinanceRequestError`/`YfinanceParseError` subclasses** — per #67 deletion-trigger discipline (filed today, trigger fired when this second service shipped with no retry-consumer). Only `YfinanceError` (catch-all base) is shipped.
- **Does NOT add `market_time` to `QuantOutput`.** β1 deferred to #68 (filed today, blocks RiskAgent #12 design phase). #10 ships α (fetch-time).
- **Does NOT do multi-horizon `dev_ibov` (1W, 1M).** Single 1-day return only per the morning-note cadence. Multi-horizon deferred to #65 (filed today, triggered when a consumer needs multi-timeframe).
- **Does NOT touch `app/utils/freshness.py`.** That file has zero production callers (verified via grep); #10's freshness check lives inline in `YfinanceService` because it uses source-side `regularMarketTime` (data-age), not fetch-latency. The dead `freshness.py` audit deferred to #66 (filed today).
- **Does NOT retroactively delete `TavilyRequestError`/`TavilyParseError` from `app/services/tavily.py`.** Per #67 (filed today — the Q2 deletion-trigger from PR #64 review fired with this PR). Cleanup is a separate chore.

## How to verify

Three gates (matches #62 verification discipline):

```bash
uv run ruff check .                                                               # full repo
uv run mypy app/services/yfinance.py app/agents/quant.py app/graph/state.py        # new + modified files
uv run pytest tests/ -v                                                           # 49 passed, 12 skipped (12 are DB-gated)
```

Specifically the 14 new tests:

```bash
uv run pytest tests/unit/test_yfinance_service.py tests/unit/test_quant_agent.py -v
# expect 14 passed
```

Live-data diagnostic (catches shape-mismatches unit tests can't):

```bash
uv run python scripts/run_quant_agent.py
# expect quant_metrics dict with real PETR4 metrics from yfinance + IBOV; flags=[]
```

## Decisions worth flagging

- **Service boundary = `app/services/yfinance.py`** mirroring `TavilyService`. yfinance-specific knowledge (`.SA` suffix, key-name normalization `trailingPE→pl`, `regularMarketTime` timestamp semantics, MultiIndex DataFrames) stays inside the service. Agents read `result.metrics.pl` and never touch yfinance internals.
- **Raise vs return across the boundary (Q1-of-PR-#64 discipline carried forward):**
  - Empty ticker → raise `YfinanceError` at construction (unrecoverable by caller; only the operator can fix bad upstream state).
  - HTTP failure / IBOV fetch failure → return `YfinanceResult.error` as WARNING (transient → retryable in the future).
  - Stale market data (>max_data_age_hours) or schema break (no expected keys present) → return FATAL (persistent → operator must investigate).
- **No `YfinanceConfigError`** because yfinance has no API key — no construction-time config failure possible. Agent's construction-block is simpler than `TavilyService` consumers (no `try/except ConfigError` around construction).
- **FATAL-on-missing-all-expected-keys, not FATAL-on-any-null-field.** Different threshold from `TavilyService`'s FATAL rule (any-null-article). Driven by cardinality: Tavily returns a list (any-null = batch breakdown by correlation); yfinance returns one metrics bundle (some-null = "metric not applicable to this ticker," all-null = "schema broke"). Operator-paging rule tracks the legitimate-missing-field pattern of each source.
- **Gate ordering: freshness check before schema-FATAL check.** Root-cause before symptom: when both conditions hold (stale data + missing keys), stale is usually the cause (yfinance serving an old cached response). Flagging the cause first sends the operator to investigate freshness, which resolves the missing-keys issue as a side effect.
- **`dev_ibov` uses `yf.Ticker(...).history()` for both ticker and IBOV** — not `yf.download(...)` for IBOV. `yf.download` returns a MultiIndex DataFrame; `yf.Ticker(...).history()` returns single-index. Same API on both sides = symmetric shape = simpler `_one_day_return` helper (no MultiIndex squeeze). Original Chunk 4 used `yf.download` for IBOV; live-script run caught the shape-mismatch (`'DataFrame' object has no attribute 'tolist'`) and we fixed it before commit.
- **`state["data_freshness"]["quant"] = datetime.now(UTC)` (Option α, fetch-time, not Option β1 data-time).** Cross-agent consistency with macro/company (all three use fetch-time). β1 (data-time via `regularMarketTime`) deferred to #68 — triggered when RiskAgent #12 design phase starts.

## Five followups filed during this PR

- **#65** — multi-horizon `dev_ibov` (1W + 1M). Trigger: RiskAgent #12 or EditorAgent needs multi-timeframe.
- **#66** — delete or revive dead `freshness.py` (zero production callers). Low-priority chore.
- **#67** — delete doc-only `TavilyRequestError` + `TavilyParseError` from `tavily.py`. The Q2 deletion-trigger from PR #64 review fired when this second service shipped with no retry-consumer.
- **#68** — add `market_time` to `QuantOutput`. Blocks RiskAgent #12 design phase. Coordinates with #65 (both touch `QuantOutput` schema).
- **#69** — canonicalize `DataFlag` imports across test files (use `app.utils.flags`, not `app.graph.state` re-export). Low-priority cleanup chore.

## Fifth mock-fraud instance found and fixed during this PR

The session surfaced a new mock-fraud member despite consistent discipline:

| # | Bug class | Surfaced by |
|---|---|---|
| 1-4 | (from PR #62) | reading / debugging / layered tests |
| **5** | `yf.download` MultiIndex vs `Ticker.history` single-index — mock returned single-index shape for `download`, real `download` returns MultiIndex. Tests passed because the mock shape didn't match the real shape. | live-script run (this PR) |

The yfinance MultiIndex bug was caught only by the live-data diagnostic script — unit tests structurally can't see shape-mismatches against real libraries, only behavior-mismatches against mocked shapes. Validates the third-verification-layer discipline: ship the live script as part of every new-service / new-agent PR. The fix (switching IBOV fetch from `yf.download` to `yf.Ticker(...).history()`) eliminated the asymmetric API at the source, and the test mock pattern was updated to `side_effect` dispatch per-ticker so the bug class can't recur.

Closes #10
Ref #62 (TavilyService precedent)
Ref #67 (deletion-trigger for TavilyRequestError/TavilyParseError)
Ref #68 (β1 market_time, blocks #12)


pradyumna-001:
Ready for review. Verification gates passed: ruff ✅, mypy ✅, 14/14 unit tests ✅, live-data script ✅. Please review and triage.

app/services/yfinance.py
```diff
@@ -0,0 +1,136 @@
...
+
+@dataclass(frozen=True)
+class YfinanceResult:
+    metrics: QuoteMetrics | None
+    error: DataFlag | None
+
+
+class YfinanceService:
+    def __init__(self, ticker: str, market_index: str = "^BVSP", max_data_age_hours: int = 24) -> None:
+        if not ticker:
```
pradyumna-001, line 34: Q1 (architecture) — app/services/yfinance.py:34-35: Why does YfinanceService raise YfinanceError on empty ticker instead of returning a WARNING result in the YfinanceResult? This follows the #62 PR #64 Q1 discipline (raise-vs-return asymmetry across the boundary). Walk me through the design rationale.

app/services/yfinance.py
```diff
@@ -0,0 +1,136 @@
...
+                        source="yfinance",
+                        severity=Severity.FATAL,
+                        message=(
+                            f"Yfinance data for {self.ticker!r} is "
+                            f"{data_age_hours:.1f}h old (max {self.max_data_age_hours}h)"
+                        )
+                    )
+                )
+
+        expected_keys = {"trailingPE", "enterpriseToEbitda", "priceToBook", "dividendYield"}
```
pradyumna-001, line 83: Q2 (architecture) — app/services/yfinance.py:83-96: The FATAL-on-missing-all-expected-keys threshold differs from TavilyService's FATAL-on-any-null rule. This decision is driven by cardinality: Tavily returns a list (any-null = batch breakdown by correlation), yfinance returns one metrics bundle (some-null = metric not applicable, all-null = schema broke). Why this different threshold, and how does the operator-paging rule track legitimate-missing-field patterns differently per source?

app/services/yfinance.py
```diff
@@ -0,0 +1,136 @@
...
+            return YfinanceResult(
+                metrics=None,
+                error=DataFlag(
+                    source="yfinance",
+                    severity=Severity.WARNING,
+                    message="Yfinance returned no info dict"
+                )
+            )
+
+        market_time = info.get("regularMarketTime")
```
pradyumna-001, line 64: Q3 (architecture) — app/services/yfinance.py:64-96: Gate ordering places freshness check before schema-FATAL check. The rationale is root-cause-before-symptom: when both stale data (>max_age) and missing keys are present, stale is usually the cause (yfinance serving old cached response). Flagging the cause first sends the operator to investigate freshness, which resolves the missing-keys issue as a side effect. Was this ordering intentional, and does it reflect the broader principle of addressing source-side freshness before schema-level symptoms?

pradyumna-001:
Q4 (architecture) — app/graph/state.py:21-27: QuantOutput field types widened from float to float|None (per the partial-success principle). A non-dividend-paying company legitimately has dividendYield: None. The | None encodes 'metric not applicable' at the type level. Was this widening driven by the same YAGNI-then-reality principle from #62, or was it a direct carry-over from the TavilyService pattern?

pradyumna-001:
Q5 (architecture) — app/services/yfinance.py:99-101,122: The IBOV history fetch uses  not . This avoids the MultiIndex shape mismatch (the fifth mock-fraud finding: real  returns MultiIndex DataFrame, tests used single-index mock). Why was  chosen over  for IBOV, and does this reflect a general principle of matching the exact API shape between mock and production code?

pradyumna-001:
## Closed-Issue Peer Review: #10 QuantAgent (PR #70)

Five architectural questions posted as inline review comments on the diff:

1. **Q1** (yfinance.py:34-35): Why raise YfinanceError on empty ticker vs return WARNING in YfinanceResult? Follows #62 PR #64 Q1 discipline.
2. **Q2** (yfinance.py:83-96): FATAL-on-missing-all-expected-keys threshold vs TavilyService's FATAL-on-any-null. Driven by cardinality differences.
3. **Q3** (yfinance.py:64-96): Gate ordering — freshness check before schema-FATAL. Root-cause-before-symptom principle.
4. **Q4** (state.py:21-27): QuantOutput float→float|None widening. Partial-success principle. YAGNI-then-reality or direct carry-over from Tavily?
5. **Q5** (yfinance.py:99-101,122): IBOV history uses  not . Avoided MultiIndex shape mismatch (5th mock-fraud finding).

I'll await the user's defenses in chat, then mirror their answers onto the PR with ✅ Solved markers.

Per the closed-issue peer-review discipline:
- Scope: PR #70's code only (no peeking at new issue)
- Time-box: 15-20 min
- No code editing during review
- Honest misses are wins
- Questions only — user posts their own defenses

app/services/yfinance.py
pradyumna-001: The raise-vs-return asymmetry from #62 Q1 was specifically about what the operator can control. With Tavily, the API key is a config-time concern - if it's missing, the service can't function, so raise at construction makes sense. With yfinance, there's no API key, so an empty ticker is purely a caller mistake. 

Raising at construction means the bug is caught immediately when the service is instantiated, rather than latent until search() runs. This keeps the pipeline lean: no invalid service objects float around, no extra None-checks in search(), and the operator gets instant feedback. It also mirrors the pattern from #62 where empty-string validation on the source side was a raise - just applied at the construction boundary instead of the search boundary because yfinance has no config-time guard (no key to check).

app/services/yfinance.py
pradyumna-001: The FATAL threshold difference comes down to what each source returns. Tavily articles come in a list - if any single article has a null title, that's a real breakdown across the batch, so FATAL-on-any-null makes sense because the operator needs to know the whole correlation result is compromissed. But yfinance returns one metrics bundle for a single ticker. If trailingPE is None, that just means this specific company doesn't pay a dividend or has different reporting format - it's 'metric not applicable,' not 'schema broke.'

Only if ALL four expected keys are missing does it signal the yfinance response format itself changed, which is an operator-pagin event. So the threshold tracks: some-null = 'ignore, metric doesn't apply to this ticker' vs all-null = 'investigate, the API may have changed.' This keeps the operator from being paged on legitimate missing fields while still catching real schema breaks.

app/services/yfinance.py
pradyumna-001: The gate ordering is explicitly root-cause-before-symptom. When yfinance serves a stale cached response, the missing keys are a symptom of the stale data - the cache didn't fetch fresh info. If we flag the schema-FATAL first, the operator investigates 'why are keys missing?' and might never realize the data is simply old.

By checking freshness first, we say 'the data is X hours old (above the max)' and the operator can refresh or investigate the cache layer, which then naturally resolves the missing-keys issue. This isn't just about yfinance - it's a broader principle: when a source has both a freshness timestamp and key presence, freshness is the root cause in the common case, so flag it first . The operator then has the option to ignore the FATAL if they've already confirmed freshness, but the reverse path (schema-first) risks hiding the freshness problem.

